PR #856, opened 2026-08-07
chore: update version 1.2.0-alpha.75

ByteZhang1024:


wabicai (approved):

packages/connect-examples/hwk-demo/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "hwk-demo",
-  "version": "1.2.0-alpha.70",
+  "version": "1.2.0-alpha.75",
```
originalix, line 3: P2: [Align hwk-demo with the alpha.75 workspace packages]

This manifest now declares alpha.75, but its eight `@onekeyfe/hwk-*` dependencies are still pinned to `1.1.34-alpha.1`. Because those exact ranges do not match the local alpha.75 workspaces, Yarn reports them as mismatched and resolves the old registry builds, so the demo can pass without exercising the packages this release is publishing.

Update the HWK dependency pins to `1.2.0-alpha.75`, refresh dependency resolution if needed, and confirm that `yarn workspaces info` no longer reports mismatches.

packages/connect-examples/expo-example/src/data/output.json
```diff
@@ -0,0 +1,2019 @@
+{
```
originalix, line 1: P2: [Remove the generated transaction dump from the release diff]

`output.json` is the nondeterministic artifact written by `bitcoin.test.ts`; it is not imported by the Expo app and contains 501 repeated transaction outputs. Because the generator uses `crypto.randomBytes`, rerunning it changes the tracked file, leaving dirty trees and adding unrelated transaction data to future release reviews.

Delete this file and direct any diagnostic output to an ignored location outside the tracked source tree.

originalix:
## Automated code review found 2 issues to address

@ByteZhang1024

Reviewed commit `1905c5a3d6a1`. This report always reflects the latest reviewed changes.

### Review summary

This release-oriented change leaves the hd-* and hwk-* architecture boundaries and device protocol behavior unchanged while advancing package manifests and most exact internal dependency pins from 1.2.0-alpha.70 to 1.2.0-alpha.75 across browser, BLE, USB, bridge, emulator, CLI, and example entry points. It also adds an unconsumed generated Bitcoin reference-transaction JSON file to the Expo example.

**What needs attention:** Align the HWK demo with the alpha.75 workspaces and remove the nondeterministic transaction dump before merging, then complete lint and build validation.

### Issues to address

- **P2 · Align hwk-demo with the alpha.75 workspace packages**
  - **Impact:** The demo manifest now declares 1.2.0-alpha.75, but its eight internal HWK dependencies remain pinned to 1.1.34-alpha.1. Yarn classifies every one as a mismatched workspace dependency, so it resolves old registry builds instead of linking the alpha.75 packages. Demo validation can therefore pass without exercising the release under review.
  - **Suggested action:** Update the eight @onekeyfe/hwk-* dependency pins to 1.2.0-alpha.75, refresh dependency resolution if necessary, and verify that yarn workspaces info reports no mismatch for hwk-demo.
- **P2 · Remove the generated transaction dump from the release diff**
  - **Impact:** The added output.json is an unconsumed artifact produced by bitcoin.test.ts. Its generator uses crypto.randomBytes, so regenerating it changes tracked content, while its 501 repeated outputs add unrelated transaction-shaped data and substantial noise to a package-version release.
  - **Suggested action:** Delete output.json and prevent the diagnostic test output from being tracked, such as by writing it outside the source tree and ignoring that location.

### Validation gaps

- `yarn check-versions` passed and reported all 30 top-level packages at 1.2.0-alpha.75.
- `yarn workspaces info` identified only `hwk-demo`, with eight mismatched internal dependencies.
- The repository's PR validation stopped at full lint because the installed dependencies do not contain the `eslint` executable; build and later checks were therefore not executed.
- No physical-device validation was performed, and none is required for the manifest-only behavior, but package builds and example type checks should complete before publication.

PR #858, opened 2026-08-08
feat: support Pro2 and Neo in SDK examples

wabicai:


packages/connect-examples/expo-example/src/views/FirmwareScreen/ProtocolV2FirmwareUpdate.tsx
```diff
@@ -0,0 +1,398 @@
...
+        payload: error instanceof Error ? error.message : String(error),
+      });
+    }
+  };
+
+  const runUpdate = async (useRemoteConfig: boolean) => {
+    setIsUpdating(true);
+    setUpdateState(undefined);
+    try {
+      const params: FirmwareUpdateV4Params = {
```
originalix, line 168: P1: [Remote update invokes firmwareUpdateV4 without targets]

Clicking Remote Update calls `runUpdate(true)`, which builds parameters containing only `platform`. `FirmwareUpdateV4` filters remote components against `targetsToUpdate`, so an empty target set downloads nothing and ends with `No firmware to update`.

Call `checkAllFirmwareRelease` for the selected device first and pass its `targetsToUpdate` into `firmwareUpdateV4`, handling an empty list as an already-current device.

packages/connect-examples/expo-example/src/components/UploadScreen/nftUtils.ts
```diff
@@ -42,6 +42,25 @@ function htmlImageToCanvas({
   return { canvas, ctx };
 }
 
+/** Convert an image source to exact-size RGBA pixels for Protocol V2 image APIs. */
+export async function imageSourceToRgba(source: string, width: number, height: number) {
+  if (typeof document === 'undefined') {
```
originalix: P1: [Protocol V2 image uploads are web-only in the Expo app]

On React Native, `document` is undefined, so this helper throws before the new Pro2/Neo wallpaper or NFT path can call the device. That makes both Protocol V2 image actions unusable over React Native BLE.

Add a React Native-compatible exact-size RGBA conversion path and select it by platform, while retaining the canvas implementation for web.

packages/connect-examples/expo-example/src/components/UploadScreen/index.tsx
```diff
@@ -322,11 +393,31 @@ function UploadScreenComponent() {
           alert('Wallpaper uploaded successfully');
         }
       } else {
-        if (!uploadResParams) {
+        if (!uploadResParams && !protocolV2NftParams) {
           alert('Please load NFT data first');
           return;
         }
 
+        if (selectedDevice?.connectProtocol === 'V2') {
```
originalix: P1: [BLE NFT upload drops the detected V2 protocol]

BLE discovery intentionally leaves `connectProtocol` unset. `loadNftData` can still actively detect V2 and prepares `protocolV2NftParams`, but this branch checks only the original discovery object, falls through to the V1 path, and then fails because `uploadResParams` was cleared.

Persist the active protocol returned during preparation and use that value here to choose `deviceUploadNft` versus `deviceUploadResource`.

packages/connect-examples/expo-example/src/views/FirmwareScreen/ProtocolV2FirmwareUpdate.tsx
originalix, line 168: P1: [Remote update invokes firmwareUpdateV4 without targets]

Clicking Remote Update builds parameters containing only `platform`. `FirmwareUpdateV4` filters remote components against `targetsToUpdate`, so the empty target set downloads nothing and ends with `No firmware to update`.

Call `checkAllFirmwareRelease` for the selected device first and pass its `targetsToUpdate` into `firmwareUpdateV4`, treating an empty list as an already-current device.

packages/connect-examples/expo-example/src/components/UploadScreen/nftUtils.ts
originalix: P1: [Protocol V2 image uploads are web-only in the Expo app]

On React Native, `document` is undefined, so this helper throws before the new Pro2/Neo wallpaper or NFT path can call the device. That makes both Protocol V2 image actions unusable over React Native BLE.

Add a React Native-compatible exact-size RGBA conversion path and select it by platform, while retaining the canvas implementation for web.

packages/connect-examples/expo-example/src/components/UploadScreen/index.tsx
originalix: P1: [BLE NFT upload drops the detected V2 protocol]

BLE discovery leaves `connectProtocol` unset. `loadNftData` can still actively detect V2 and prepares `protocolV2NftParams`, but this branch checks only the original discovery object, falls through to the V1 path, and then fails because `uploadResParams` was cleared.

Persist the active protocol returned during preparation and use that value here to choose `deviceUploadNft` versus `deviceUploadResource`.

packages/connect-examples/expo-example/src/views/FirmwareScreen/ProtocolV2FirmwareUpdate.tsx
wabicai: Addressed in commit 1d555096b. The remote action now calls checkAllFirmwareRelease for the selected device and runtime platform, passes the returned targetsToUpdate into firmwareUpdateV4, and short-circuits with an already-current success message when the plan is empty. Verified with Expo TypeScript checks, ESLint, focused tests, and the repository commit check.

wabicai: Addressed in commit 1d555096b. The remote action now calls checkAllFirmwareRelease for the selected device and runtime platform, passes the returned targetsToUpdate into firmwareUpdateV4, and short-circuits with an already-current success message when the plan is empty. Verified with Expo TypeScript checks, ESLint, focused tests, and the repository commit check.

packages/connect-examples/expo-example/src/components/UploadScreen/nftUtils.ts
wabicai: Addressed in commit 1d555096b. imageSourceToRgba now selects a React Native path outside web: it materializes remote/data sources, performs cover resize plus exact cropping with expo-image-manipulator, decodes the result into exact-size RGBA bytes, and cleans temporary files. The existing Canvas implementation remains unchanged for web. Added focused tests for crop geometry and native RGBA conversion.

wabicai: Addressed in commit 1d555096b. imageSourceToRgba now selects a React Native path outside web: it materializes remote/data sources, performs cover resize plus exact cropping with expo-image-manipulator, decodes the result into exact-size RGBA bytes, and cleans temporary files. The existing Canvas implementation remains unchanged for web. Added focused tests for crop geometry and native RGBA conversion.

packages/connect-examples/expo-example/src/components/UploadScreen/index.tsx
wabicai: Addressed in commit 1d555096b. NFT preparation now persists the actively detected protocol in preparedNftProtocol, and upload dispatch uses that prepared value instead of the discovery objects optional connectProtocol. Prepared params/protocol are cleared when the selected connectId changes or a reload starts, and the V2 path validates its connectId before deviceUploadNft.

wabicai: Addressed in commit 1d555096b. NFT preparation now persists the actively detected protocol in preparedNftProtocol, and upload dispatch uses that prepared value instead of the discovery objects optional connectProtocol. Prepared params/protocol are cleared when the selected connectId changes or a reload starts, and the V2 path validates its connectId before deviceUploadNft.

packages/core/src/device/DeviceStateStore.ts
```diff
@@ -148,6 +148,13 @@ export class DeviceStateStore {
     if (patch.status) applySectionPatch(next.status, patch.status, 'status', changedKeys);
     if (patch.settings) applySectionPatch(next.settings, patch.settings, 'settings', changedKeys);
     if (patch.versions) applySectionPatch(next.versions, patch.versions, 'versions', changedKeys);
+    if (
+      patch.securityElements !== undefined &&
+      !isEqual(next.securityElements, patch.securityElements)
+    ) {
+      next.securityElements = cloneDeviceState(patch.securityElements);
```
originalix: P2: [Partial secure-element patches replace the entire collection]

`securityElements` is a partial record, but this assignment replaces the whole stored collection. When a later state refresh contains only one secure element, metadata already read for the other elements is discarded and the firmware-information projection becomes incomplete.

Merge by secure-element key (and by `type`/`state` within each entry), preserving entries that are absent from the patch.

wabicai: Addressed in commit 9e7f78221. DeviceStatePatch.securityElements is now explicitly modeled as a per-element partial patch. DeviceStateStore merges first by SE key and then by type/state within each entry, preserving secure elements and fields omitted from incremental refreshes. Added a regression test that updates only se01.state after storing SE01 and SE02, and verifies both the canonical state and projected compatibility fields remain complete. Core TypeScript, 12 focused tests, the full hd-core test suite/build, lint, and the repository commit check all pass.

originalix (approved):

PR #862, opened 2026-08-10
fix: pause bootloader timeout for webusb authorization

originalix:
## Summary

- pause the shared bootloader reconnect deadline while Chrome is waiting for WebUSB authorization
- restart bootloader polling with a fresh deadline after the user authorizes the re-enumerated device
- add regression coverage for the WebUSB bootloader authorization handoff
- release the aligned SDK packages as `1.2.0-alpha.97`

## Root cause

When a Protocol V1 WebUSB device reboots into bootloader mode, Chrome requires the bootloader device to be authorized again. `FirmwareUpdateBaseMethod` awaited that browser permission flow without clearing its existing polling interval and timeout. The timeout could therefore reject with `DeviceNotFound` while the WebUSB prompt was still pending, even if the user subsequently authorized the device.

The fix clears the active polling timers before requesting authorization. Once authorization completes, the existing polling loop starts again with a fresh timeout budget.

## Compatibility

- Protocol V1 V3 WebUSB bootloader flow: fixed
- FirmwareUpdateV2: unchanged; it uses an independent polling implementation
- legacy V1 updater: unchanged; it uses an independent polling implementation
- FirmwareUpdateV4 / Protocol V2: unchanged; it uses an independent reconnect path

## Validation

- firmware update test suite: 13 suites / 94 tests passed
- targeted V2, V3, and legacy bootloader polling tests passed
- V3 reconnect regression tests passed
- Protocol V2 / V4 isolation coverage passed
- targeted ESLint and `packages/core` build passed
- `yarn agent:check --profile commit` passed
- package publish and web release workflows for `1.2.0-alpha.97` succeeded


packages/core/src/api/firmware/FirmwareUpdateBaseMethod.ts
```diff
@@ -191,6 +191,8 @@ export class FirmwareUpdateBaseMethod<Params> extends BaseMethod<Params> {
           !hasPromptedWebDevice &&
           !isPromptingWebDevice
         ) {
+          clearInterval(intervalTimer);
```
originalix, line 194: P2: [WebUSB authorization can leave firmware updates pending forever]

When authorization returns without a device ID, or the first post-authorization scan does not find exactly one bootloader device, both the polling interval and deadline have already been cleared. `_checkDeviceInBootloaderMode()` then returns without settling `checkPromise`, so the firmware-update workflow can wait indefinitely and keep later device work blocked.

Please model this as a pause/resume transition: reject explicit cancellation, and after successful authorization restart bounded polling with a renewed deadline until discovery succeeds or fails.

originalix:

@originalix

Reviewed commit `d68cfec311df`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes Core-owned WebUSB firmware reboot orchestration so the polling interval and 30-second deadline stop when browser authorization starts, adds a fake-timer regression test for FirmwareUpdateV2 and FirmwareUpdateV3, and synchronizes monorepo package versions and internal dependencies to 1.2.0-alpha.97. It does not change transport framing, protocol detection, or network boundaries.

**What needs attention:** Ensure the WebUSB authorization pause has a bounded completion path when authorization is cancelled or the first post-authorization device scan does not find exactly one bootloader device.

### Issues to address

- **P2 · WebUSB authorization can leave firmware updates pending forever**
  - **Impact:** When the authorization response is empty, or the first scan after authorization returns zero or multiple bootloader devices, the polling interval and timeout have already been cleared. The scan returns without settling checkPromise, and no polling or deadline is restarted, so the firmware-update call can remain pending indefinitely and retain the in-progress device workflow.
  - **Suggested action:** Model authorization as a pause/resume transition: reject an explicit cancellation, and after successful authorization restart bounded bootloader polling until discovery succeeds or a renewed deadline expires. Add coverage for cancellation and an unsuccessful first scan.

### Validation gaps

- Package version alignment passed for all 30 checked packages.
- Diff whitespace validation passed.
- The focused Jest file could not execute because workspace dependencies are not installed and the `jest` binary is absent.
- No physical WebUSB authorization, cancellation, multiple-authorized-device, or delayed bootloader enumeration was exercised.

wabicai (approved):

PR #866, opened 2026-08-10
fix: harden Pro2/Neo recovery and resource uploads

wabicai:
## 变更概览

本 PR 汇总处理 Pro2 / Neo 固件恢复、Protocol V2 资源上传、TON 返回值与配套 SDK 发布：

- 固件升级取消时立即失效 Protocol V2 会话和运行时缓存；显式刷新 loader 状态时重新协商 ProtocolInfo。
- 为旧版 loader 的 Legacy ProtocolInfo 增加传输层解码、受限恢复入口、固件预检与 V4 更新回退，不放宽普通业务调用。
- 旧 App 不支持 `DeviceFirmwareUpdateRequest` 并重启至 bootloader 后，每次重试都重新获取当前 `DeviceCommands`，避免复用已失效的命令通道；回归测试会在重启时替换命令对象。
- 统一 TON 消息签名返回结构：兼容历史 `signning_message`，对外输出 `signing_message`。
- 回滚 TopLevel / LowLevel 全局二进制 codec，避免改变固件、Bootloader 和普通 SDK 调用的通用传输语义。
- Protocol V2 壁纸、NFT、Portfolio API 改用 JSON-safe Base64 参数；严格 Base64/JPEG 校验、JPEG 解码、RGBA/设备格式转换与分块写入统一下沉到 SDK LowLevel。
- 壁纸上传补齐消息能力判断；不支持 FileWrite / DirMake / DeviceSettingsSet 的固件会在写入前返回明确错误。
- CLI、Expo 示例与 Playground 同步使用新的 Base64 API。
- 本次新增的中文代码注释已统一改为英文。
- 全部工作区升级为 `1.2.0-alpha.104`，并通过官方工作流发布到 npm `next`。

## 设计结果

- Extension、React Native、Electron 与 Web 只跨运行时传递字符串和普通对象，不依赖 ArrayBuffer / TypedArray 的私有桥接恢复。
- JPEG 只在 SDK 内解码一次；设备尺寸、RGBA 长度、文件大小和消息能力由 SDK 统一校验。
- 固件 V1–V4、Bootloader、完整资源升级与 Protocol V1 资源上传保持原有边界和行为。
- App 与 SDK 同步升级，无需保留旧参数的双写兼容层。

## 验证

- 最新 PR CI：lint、build、CodeQL、Socket 与 Snyk 全部通过。
[...]

PR #867, opened 2026-08-10
fix: preserve binary payloads across SDK runtime bridges

wabicai:
## What changed

- recursively encode `ArrayBuffer`, typed-array views, and `Blob` inputs at the TopLevel/LowLevel boundary
- restore tagged Base64 payloads before SDK method construction and validation
- preserve `ArrayBuffer` versus `Uint8Array` semantics and exact typed-array slice boundaries
- reject malformed Base64, malformed tagged objects, and circular payloads before hardware I/O
- document the cross-runtime binary boundary
- add public API regression coverage for Portfolio, wallpaper, NFT, firmware V1–V4, bootloader, and full resource uploads

## Why

TopLevel and LowLevel SDK instances can be separated by JSON-only hosts such as extension background/offscreen bridges. Native binary values lose their runtime type during serialization, which caused `FilesystemFileWrite` validation to reject Portfolio data before reaching the device.

## Impact

SDK consumers receive a generic, host-independent binary transport boundary. Existing same-runtime LowLevel callers, protobuf payloads, filesystem formats, and device firmware protocols are unchanged.

Companion App PR: https://github.com/OneKeyHQ/app-monorepo/pull/12814

## Validation

- `bridgeBinaryPayload.test.ts`: 5 tests passed
- targeted ESLint passed
- `git diff --check` passed
- no hardware or firmware command was executed

### DIFF
--- a/docs/architecture/decisions.md
+++ b/docs/architecture/decisions.md
@@ -51,6 +51,25 @@ Pro2，也不能用 Pro/Pro2 型号反推协议。这样 Pro 后续迁移到 Pro
 - `packages/hd-transport/src/protocols/v2/frame-assembler.ts`
 - `packages/hd-transport/src/protocols/v2/link-manager.ts`
 
+## TopLevel / LowLevel 跨运行时二进制边界
+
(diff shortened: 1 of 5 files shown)


wabicai:
二进制桥接修复已追加到 #866，本 PR 关闭以避免重复。

PR #869, opened 2026-08-12
fix: prevent desktop BLE background reconnects

wabicai:
## Summary

- propagate an optional `reuseConnectedOnly` flag from Core through `DeviceConnector` into Electron BLE transport
- require an already-connected, fully initialized Noble link in the Electron main process; never scan, reconnect, rediscover services, or retry in this mode
- fail closed when the desktop preload does not expose the connected-only capability, normalize IPC failures to terminal `DeviceNotFound`, and preserve the user's kept-alive BLE link when a background setup attempt fails
- stop both inner retry and outer polling after the single connected-only attempt

## Why

Desktop Portfolio Sync runs as a best-effort idle task. It must not turn a stale BLE lease into active scanning/reconnection, block foreground hardware work, or disconnect the user's active BLE session.

## Validation

- `yarn jest --runInBand` on 5 focused suites: 74 tests passed
- ESLint passed on all 18 changed TypeScript files
- `tsc` passed for `hd-transport` and `hd-transport-web-device` after building the Electron transport declarations
- Electron transport build passed; its standalone `tsc` still reports the pre-existing unused `webContents` parameter at `noble-ble-handler.ts:423`

## Integration note

The dependent App PR must remain Draft until these SDK packages are published together and the App upgrades `hd-core`, `hd-transport-electron`, and `hd-transport-web-device` from `1.2.0-alpha.108`.

### DIFF
--- a/packages/connect-examples/electron-example/src/preload.ts
+++ b/packages/connect-examples/electron-example/src/preload.ts
@@ -8,6 +8,7 @@ import { ipcMessageKeys } from './config';
 import type {
   DesktopAPI as BaseDesktopAPI,
   NobleBleAPI,
+  NobleBleConnectOptions,
   NobleBleWriteOptions,
 } from '@onekeyfe/hd-transport-electron';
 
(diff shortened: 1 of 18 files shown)


wabicai:
Closing this SDK-side approach. We are handling desktop BLE Portfolio connected-only behavior in app-monorepo instead, so these cross-layer SDK lifecycle changes are no longer needed.

PR #870, opened 2026-08-12
fix: align Pro2 firmware update flow

wabicai:
## 变更内容

- 对齐 firmware-pro2 最新 dev protobuf 与 DeviceInfo/main_mcu 字段命名。
- 按最新固件流程执行 DeviceFirmwareUpdateStage → 空 DeviceFirmwareUpdateRequest → 主动状态轮询。
- 为 Protocol V2 transport 增加写完即返回能力，避免等待安装请求同步 ACK。
- 放宽本地资源 ZIP：支持外层目录、额外文件以及无 manifest.json 的本地包。
- Pro2 Update 页连接 WebUSB 时先触发系统设备选择框，与 Playground 其他连接入口保持一致。
- 更新所有 SDK 包版本至 1.2.0-alpha.109。

## 根因

SDK 仍按旧协议执行固件升级并严格要求本地 ZIP manifest。Pro2 Update 页还调用了普通 searchDevices()，只枚举已有授权设备，没有触发浏览器 WebUSB 系统选择框。

## 验证

- Core 回归：410 项通过，4 项跳过。
- 固件与资源聚焦回归：295 项通过，4 项跳过。
- hd-transport Protocol V2：56 项通过。
- Pro2 Update 页面 ESLint 通过。
- Playground hardwareService 聚焦测试：10 项通过。
- hd-transport、hd-core 构建通过。
- 30 个包版本均为 1.2.0-alpha.109。
- GitHub Actions 完整构建、NPM 发布与 Examples 部署通过；最新提交的 PR 检查正在重新执行。


originalix:

@wabicai

Reviewed commit `eb727a93ba19`. This report always reflects the latest reviewed changes.

### Review summary

The PR migrates the Pro2/Neo Protocol V2 firmware path from a single update request to a staged-target plus empty-trigger protocol, adds write-only transport completion and revised install polling, aligns generated device-information fields, broadens local resource archive ingestion, adjusts onboarding unlock behavior, and updates example progress handling, documentation, tests, and package versions.

**What needs attention:** Review the firmware trigger boundary across legacy and current Pro2 loaders, especially response-stream ownership after the write-only trigger on WebUSB and BLE.

### Issues to address

- **P1 · Preserve legacy Pro2 firmware update compatibility**
  - **Impact:** On a loader using the previous Protocol V2 IDs, message 61000 is the old side-effecting DeviceFirmwareUpdateRequest rather than DeviceFirmwareUpdateStage. Awaiting Stage success can therefore start installation and then reject when the device reboots or BLE disconnects, preventing the high-level flow from entering its reconnect and status-polling phase even though device mutation has begun.
  - **Suggested action:** Restore an active-runtime capability or legacy ProtocolInfo branch: use the old request and expected-disconnect handling for legacy loaders, and use Stage plus the empty trigger only when the connected firmware advertises the new handlers.
- **P1 · Do not leave the fire-and-forget response in the active link**
  - **Impact:** returnAfterWrite ends the serialized Protocol V2 call immediately after native write completion, before reading the request's ACK or terminal frame. A subsequent status query or SDK command can consume that stale frame under a different request sequence, causing an ACK-sequence failure, link invalidation, or misattribution of the old terminal response.
  - **Suggested action:** Give the trigger a protocol-aware completion mode that retains ownership of the response stream until it consumes the matching ACK, then returns without waiting for the terminal business response.

### Validation gaps

- `git diff --check` completed without whitespace errors.
- The changed production files did not introduce a new outbound request or host.
- Focused Jest suites could not run because workspace test dependencies are unavailable (`jest: command not found`).
- The checked-out firmware submodule content is not populated, so device-side behavior could not be independently exercised from this workspace.
- Required validation should cover current and legacy Pro2 loader schemas on WebUSB and BLE, including ACK-before-response, disconnect-before-response, stale status records, and the first command after the install trigger.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2734,7 +2692,7 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         try {
           const statusResponse = await this.device.getCommands().typedCall(
             'DeviceFirmwareUpdateStatusGet',
-            'DeviceFirmwareUpdateStatus',
+            ['DeviceFirmwareUpdateStatus', 'Success'],
```
originalix, line 2666: P1: [Do not treat the install response as status completion]

When `DeviceFirmwareUpdateRequest` returns after writing, its eventual `Success` remains unread. This status call then accepts that same response type, so it can consume the install response and mark the update complete without target records, a verified version change, or confirmed post-install state.

Require this query to terminate only on `DeviceFirmwareUpdateStatus`; let the session skip an unrelated `Success`, and keep completion tied to target status or verified post-reboot version evidence.

packages/core/src/protocols/protocol-v2/resources.ts
```diff
@@ -140,9 +131,9 @@ function parseProtocolV2ResourceManifestFile(
     device_path: devicePath,
     size: Number(file.size),
     sha256: digest,
-    signed: true,
-    sig_algo: file.sig_algo,
-    payload_version: file.payload_version ?? null,
+    ...(file.signed === undefined ? {} : { signed: file.signed }),
```
originalix, line 134: P2: [Reject explicitly unsafe resource signing metadata]

When a manifest provides `signed: false` or an unknown `sig_algo`, this parser preserves the value and accepts the archive. That allows metadata explicitly declaring an unsigned or unsupported package to pass host preflight and reach device I/O.

Keep omission compatible with the revised manifest format, but require `signed === true` and a supported algorithm whenever those fields are present.

originalix (dismissed):

originalix (commented):
@wabicai

Reviewed commit `4223c64305d5`.

- **P1 · Do not treat the install response as status completion**
- **P2 · Reject explicitly unsafe resource signing metadata**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/870#issuecomment-5266941983)

originalix (commented):
@wabicai

Reviewed commit `eb2eb4935f9c`.

- **P1 · Do not treat the install response as status completion**
- **P1 · Migrate the low-level update method to Stage and Request**
- **P2 · Reject explicitly unsafe resource signing metadata**
- **P2 · Keep Pro2 progress remapping scoped to the Pro2 flow**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/870#issuecomment-5266941983)

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/hd-transport/src/types/messages.ts
```diff
@@ -4843,11 +4843,14 @@ export type DeviceFirmwareTarget = {
...
 
-// DeviceFirmwareUpdateRequest
-export type DeviceFirmwareUpdateRequest = {
+// DeviceFirmwareUpdateStage
+export type DeviceFirmwareUpdateStage = {
   targets: DeviceFirmwareTarget[];
 };
 
+// DeviceFirmwareUpdateRequest
+export type DeviceFirmwareUpdateRequest = {};
```
sidmorizon, line 4852: # [🟠 P1] `DeviceFirmwareUpdate.ts` 漏改:`targets` 被静默丢弃后仍发送空 install 触发命令

## 问题

`DeviceFirmwareUpdateRequest` 在本 PR 中变成空消息(`fields: {}`),但 `packages/core/src/api/protocol-v2/DeviceFirmwareUpdate.ts` 的 `run()` 未随之迁移,仍然调用:

```ts
await this.device.commands.typedCall(
  'DeviceFirmwareUpdateRequest',
  PROTOCOL_V2_FIRMWARE_UPDATE_RESPONSE_TYPES,
  { targets },
  ...
);
```

`hd-transport` 的编码路径(`serialization/protobuf/encode.ts` 的 `patch()`)只遍历 `Message.fields`,空消息会把 `targets` 静默丢弃;TS 对目标类型 `{}` 也不做多余属性检查,编译与运行都不报错。最终上链路的是一条不带任何参数的 `DeviceFirmwareUpdateRequest`(新编号 `61001`),且没有先发 `DeviceFirmwareUpdateStage`。`protocol-v2.test.ts` 里该方法的测试 mock 了 `typedCall`,仍在断言 `targets` 传参,因此测试全绿但线上行为已改变。

## 影响

在新固件语义下,空 `DeviceFirmwareUpdateRequest` = "安装当前已 staged 的目标"。该方法一旦被调用(它是 `FirmwareUpdateV2.ts:505` 注释指向的 Pro2 低层更新路径,虽然当前未注册进公开 method registry),会完全忽略调用方指定的 `target_id`/`path`,直接触发设备安装此前残留的 staged 内容——一个副作用命令带着错误的参数语义静默执行。若设备无 staged 内容则得到与参数无关的 Failure,排查方向也会被误导。

## 建议

将该方法迁移到新流程,或在本 PR 中明确移除它:

```ts
await commands.typedCall('DeviceFirmwareUpdateStage', 'Success', { targets });
await commands.call('DeviceFirmwareUpdateRequest', {}, { returnAfterWrite: true });
```

同时把 `protocol-v2.test.ts` 中相应断言改为验证 Stage 携带 `targets`,避免 mock 掩盖编码层丢参。

[Claude session ↗](https://claude.ai/code/session_01H7cELzz7c1Kby3jN3Dc1bD)

---

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1397,15 +1354,21 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
           HardwareErrorCode.RuntimeError,
           `Protocol V2 local resource file does not match manifest: ${file.archive_path}`,
           { firmwareUpdateCode: 'FirmwareArtifactReceiptMismatch' }
         );
       }
+      normalizedFiles.push({ ...file, sha256: digest });
       materializedEntries.push({ entryName: file.archive_path, binary: fileBinary });
     }
+    manifestBinary ??= new TextEncoder().encode(JSON.stringify({ files: normalizedFiles })).buffer;
+    materializedEntries.unshift({ entryName: 'manifest.json', binary: manifestBinary });
```
sidmorizon, line 1372: 🟠 P1: Firmware resource updates downloaded from the release server can be rejected as tampered

Severity: `severe`

The resource package contents extracted for verification are now named and filtered differently than the ones the host application recorded when it downloaded the package (`materializedEntries.push({ entryName: file.archive_path, ... })` at `packages/core/src/api/FirmwareUpdateV4.ts:1367-1371`), so a resource package that contains a wrapper folder or any extra file is treated as not matching its approved download and the update aborts.
Impact: Users updating with the newly supported resource package layout over the network get a hard failure that looks like a tampering error instead of a completed update.

<details>
<summary>Entry-name/filter divergence between core extraction and host receipt</summary>

`prepareProtocolV2LocalResourceArchive` now (a) strips the manifest's directory prefix, using `file.archive_path` as `entryName`, and (b) only materializes `manifest.json` plus the manifest-selected `.okpkg` files, silently ignoring extra entries such as `build-info.txt` (see the new test at `packages/core/__tests__/protocol-v2.test.ts:7186-7189`).

The prepared-plan path in `packages/core/src/api/FirmwareUpdateV4.ts:1438-1457` re-extracts the archive with this same function and then requires **exact set equality** with `archiveArtifact.materializedEntries`:

```ts
if (
  entries.length !== verifiedArchive.materializedEntries.length ||
  verifiedArchive.materializedEntries.some(entry => { ... })
) throw ... 'prepared resource entries do not match the approved archive'
```

For the local `resourceArchiveBinary` flow this is self-consistent, because the plan is built from core's own output (`packages/core/src/api/FirmwareUpdateV4.ts:1180`). For the remote flow the receipt is produced by the host helper, which materializes **every** non-directory entry under its **full** ZIP name (`packages/connect-examples/expo-playground/app/utils/firmwareUpdatePlanHost.ts:66-75`, `entryName: entry.name`). With the nested `pro2-resource/...` layout plus `build-info.txt`, host entries are `pro2-resource/manifest.json`, `pro2-resource/bundles/...`, `pro2-resource/build-info.txt` while core produces `manifest.json`, `bundles/...` — both the count and the names differ, so the comparison always fails.

Either the host receipt builder must apply the same prefix-stripping/filtering, or the comparison must be relaxed to "every verified entry is present and matching" instead of exact set equality.

</details>

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2734,7 +2689,7 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         try {
           const statusResponse = await this.device.getCommands().typedCall(
             'DeviceFirmwareUpdateStatusGet',
-            'DeviceFirmwareUpdateStatus',
+            ['DeviceFirmwareUpdateStatus', 'Success'],
```
originalix, line 2666: P1: [Preserve response ownership after the write-only install request]

When `DeviceFirmwareUpdateRequest` returns immediately after writing, its eventual `Success` remains in the Protocol V2 receive stream. If it arrives after the reconnect identity probe, this status call can consume that prior response and mark the update complete without requested-target records, a verified version change, or confirmed post-install state.

Do not let a bare `Success` complete polling; require requested-target status or validated post-reboot runtime and version evidence.

packages/hd-transport/src/types/messages.ts
originalix, line 4852: P1: [Migrate the low-level update method to Stage and Request]

Existing `deviceFirmwareUpdate` callers still pass their target list in `DeviceFirmwareUpdateRequest`. With this message now empty, protobuf encoding discards those targets and the Core method never sends `DeviceFirmwareUpdateStage`, so the operation can fail or act on previously staged state.

Update the method to send `DeviceFirmwareUpdateStage` with the normalized targets, await its acceptance, and then issue this empty Request with the intentional write-only behavior.

packages/connect-examples/expo-example/src/views/FirmwareScreen/firmwareUpdateMessages.ts
```diff
@@ -1,4 +1,20 @@
...
+  progress,
+  progressType,
+}: {
+  previous: number;
+  progress: number;
+  progressType: IFirmwareUpdateProgressType;
+}) {
+  const normalized = Math.max(0, Math.min(100, progress));
+  if (progressType === 'installingFirmware') {
+    return normalized >= 100 ? 100 : Math.max(previous, 99);
```
originalix, line 14: P2: [Keep Pro2 progress remapping scoped to the Pro2 flow]

When a Protocol V1 or earlier update emits `installingFirmware` at 0 or an intermediate percentage, this unconditional branch immediately displays 99%. The shared listener handles those flows too, so their real installation progress is hidden and a stalled update appears nearly complete.

Pass device and protocol context into this helper and apply the 99% bridge only to Pro2 or Neo Protocol V2 V4 updates; preserve normalized native progress otherwise.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2744,22 +2703,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
+            !requireCurrentInstallStatus ||
+            currentInstallStatusObserved ||
+            !allReportedTargetsFinished;
+          if (
+            shouldVerifyTargetCompletion &&
+            this.assertProtocolV2TargetStatus(statusTargets, expectedTargetIds, expectedPaths)
+          ) {
             this.protocolV2FinalStatusVerified = true;
             return;
           }
```
sidmorizon: **A firmware install that finishes quickly can be reported as failed or hang for minutes**

Severity: `non-severe`

The update is only accepted as done when the device is first seen actively installing (`currentInstallStatusObserved` gate at `packages/core/src/api/FirmwareUpdateV4.ts:2745-2755`), so an install that is already finished when the first status check happens is treated as leftover data from an earlier run and never accepted.
Impact: A successful firmware update can keep spinning for up to eight minutes and then surface as an error to the user.

<details>
<summary>Stale-record detection cannot distinguish a just-finished install from an old one</summary>

`waitForProtocolV2FirmwareUpdateComplete` is always called with `requireCurrentInstallStatus = true` (`packages/core/src/api/FirmwareUpdateV4.ts:2408-2410`, after only a 1s `PROTOCOL_V2_INSTALL_STATUS_INITIAL_DELAY`). `currentInstallStatusObserved` is set only when a matching record is PENDING or IN_PROGRESS (`packages/core/src/api/FirmwareUpdateV4.ts:2716-2726`). If the first poll already returns every requested target as FINISHED, `allReportedTargetsFinished` is true and `shouldVerifyTargetCompletion` becomes false, so `assertProtocolV2TargetStatus` is never consulted and the loop keeps polling with the "status is stale" error.

Recovery then depends on secondary evidence: the empty-records + normal-mode branch requires `installEvidenceObserved` (also gated on `currentInstallStatusObserved`) or `hasProtocolV2InstallVersionChanged`. When the installed target has no observable version (e.g. SE/bootloader-only updates) the loop runs to `PROTOCOL_V2_INSTALL_TIMEOUT` (8 minutes) and rejects with a RuntimeError even though the install succeeded.

</details>

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2744,22 +2703,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
+          const allReportedTargetsFinished =
+            matchingStatusTargets.length > 0 &&
+            matchingStatusTargets.every(target => isProtocolV2TargetStatusFinished(target.status));
+          const shouldVerifyTargetCompletion =
+            !requireCurrentInstallStatus ||
+            currentInstallStatusObserved ||
+            !allReportedTargetsFinished;
+          if (
+            shouldVerifyTargetCompletion &&
+            this.assertProtocolV2TargetStatus(statusTargets, expectedTargetIds, expectedPaths)
```
sidmorizon, line 2718: # [🟠 P1] 残留的 FAILED 记录绕过 staleness 门控,重试安装被上一轮错误立即终止

## 问题

新增的 `shouldVerifyTargetCompletion` 门控只在 `allReportedTargetsFinished` 时屏蔽陈旧记录:

```ts
const shouldVerifyTargetCompletion =
  !requireCurrentInstallStatus ||
  currentInstallStatusObserved ||
  !allReportedTargetsFinished;
```

本 PR 的测试("ignores stale finished records until the current install starts")已确认固件模型:上一轮会话的记录会一直保留到新安装真正开始。但若上一轮失败,残留记录状态是 `FAILED_*`(`status >= 3`),不满足 `isProtocolV2TargetStatusFinished` → `allReportedTargetsFinished === false` → 门控放行,`assertProtocolV2TargetStatus` 在第一次轮询(`Stage` + fire-and-forget `Request` 后仅 `1s`,设备大概率还停在确认弹窗)就命中 `isProtocolV2TargetStatusFailed` 分支,抛出 `HardwareErrorCode.FirmwareError`;该错误属于 `isProtocolV2TerminalInstallStatusError`,两层 catch 都直接 rethrow,整个更新立即终止。

## 影响

失败后重试是最常见的恢复路径,而此路径必然携带残留 FAILED 记录:重试会在新安装开始前被上一轮的旧错误"秒杀",且 `currentInstallStatusObserved` 永远没机会置位。更糟的是设备侧流程仍在继续——用户在设备上确认后固件照常安装新 staged 目标,而 host 已报告失败并退出,造成 SDK 与设备状态分叉(与 stale-FINISHED 只是超时误报不同,这里是即时误报 + 副作用继续执行)。

## 建议

把 staleness 屏蔽对称地扩展到 FAILED 记录:`requireCurrentInstallStatus && !currentInstallStatusObserved` 时,对含 FAILED 的记录集也跳过 `assertProtocolV2TargetStatus`,归入既有的 "status is stale; waiting for the current install to start" 等待分支;仅在观察到本轮会话证据(记录先出现 `PENDING/IN_PROGRESS`,或记录集相对首次快照发生复位/变化)后才把 FAILED 判为终态。若固件能在 `DeviceFirmwareUpdateStage` 成功时清空旧记录,则可从根上消除两类 stale 分支。

[Claude session ↗](https://claude.ai/code/session_01H7cELzz7c1Kby3jN3Dc1bD)

---

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -3148,65 +3153,12 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
-        }
-      } else {
-        throw error;
-      }
-    }
-    this.protocolV2LegacyDirectUpdate = false;
-    // A Success ACK or an install-time disconnect both end confirmation. In the
-    // latter case only status polling may establish the final outcome.
+    const commands = this.device.getCommands();
+    await commands.typedCall('DeviceFirmwareUpdateStage', 'Success', { targets });
```
originalix, line 3133: P1: [Preserve the legacy Pro2 firmware-update wire flow]

When a Pro2 is running the previous application, bootloader, or romloader registry, IDs 61000, 61001, and 61002 mean update request, status get, and status response. This new sequence uses those same IDs for stage, update request, and status get, so an older loader can begin installing during the stage call and disconnect before its expected response, while the following requests are interpreted as different messages. That breaks the upgrade path for devices using the earlier registry.

Select the current or legacy firmware-update schema and command sequence from the active `ProtocolInfo` capabilities, retain the old disconnect-safe polling behavior for legacy loaders, and cover both USB and BLE upgrade paths.

packages/connect-examples/expo-example/src/views/FirmwareScreen/firmwareUpdateMessages.ts
```diff
@@ -1,4 +1,20 @@
...
+  previous,
+  progress,
+  progressType,
+}: {
+  previous: number;
+  progress: number;
+  progressType: IFirmwareUpdateProgressType;
+}) {
+  const normalized = Math.max(0, Math.min(100, progress));
+  if (progressType === 'installingFirmware') {
```
originalix, line 13: P2: [Do not collapse every install phase to 99%]

This event handler is shared by legacy firmware flows as well as Pro2 V4. When any of those flows emits `installingFirmware` below 100, this branch returns at least 99, so installation appears to jump immediately to 99% and the real progress values from V3 and other flows are discarded.

Apply this transfer/install remapping only when the event belongs to the Pro2 V4 flow, and preserve the normalized installation percentage for other firmware methods.

originalix (commented):
@wabicai

Reviewed commit `17d66de7592c`.

- **P1 · Preserve the legacy Pro2 firmware-update wire flow**
- **P1 · Accept terminal status when a fast install skips observable intermediate states**
- **P2 · Do not collapse every install phase to 99%**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/870#issuecomment-5266941983)

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -3148,65 +3108,12 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
+    await commands.call('DeviceFirmwareUpdateRequest', {}, { returnAfterWrite: true });
```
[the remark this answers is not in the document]
wabicai: 这个兼容风险成立，但当前 ProtocolInfo 只暴露消息 ID 列表；新旧 registry 复用了 61000–61002，无法据此可靠区分线序。这个 PR 不能通过猜版本安全选择旧/新 handshake。当前改动按已协调的新固件 registry 实现；旧 loader 兼容需要固件提供明确 capability/schema revision 后单独落地，因此此线程先保留开放。

packages/hd-transport/src/protocols/v2/session.ts
```diff
@@ -271,6 +272,10 @@ export class ProtocolV2Session {
         logger?.error?.(`${logPrefix} write metrics callback failed: ${String(error)}`);
       }
 
+      if (callOptions.returnAfterWrite) {
+        return { type: 'WriteCompleted', message: {} };
```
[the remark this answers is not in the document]
wabicai: 这里有潜在响应归属风险，但直接在 returnAfterWrite 中继续读取会破坏其用于设备立即重启/断链命令的语义。当前固件升级轮询不会把裸 Success 当作完成：它只记录 ACK evidence，仍要求目标状态、版本变化或已确认的重启后运行态。由于 transport 层通用清理策略尚未确定，此线程先保留开放。

packages/core/src/protocols/protocol-v2/unlockPolicyRunner.ts
```diff
@@ -80,6 +80,7 @@ export async function runMethodWithUnlockPolicy<T = unknown>(
   const requiresPreUnlock =
     device.isProtocolV2() &&
     (method.useDevicePassphraseState || method.unlockPolicy === 'unlock-before-run') &&
+    device.state?.status.initialized !== false &&
```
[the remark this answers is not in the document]
wabicai: 这里是预期行为：initialized=false 表示设备尚未初始化，不存在可执行的 PIN 解锁流程，因此所有 Protocol V2 方法都应跳过 pre-unlock，最终是否允许调用由设备端方法策略决定。把它限制到固件方法反而会让其他 pre-onboarding 管理方法错误进入 PIN 流程。

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -596,7 +563,7 @@ export const assertProtocolV2ReconnectIdentity = (
  *
  * It intentionally does not fall back to FirmwareUpdateV3/V1 behavior:
  * - upload uses FilesystemFileWrite
- * - install uses DeviceFirmwareUpdateRequest
+ * - install uses DeviceFirmwareUpdateStage followed by an empty DeviceFirmwareUpdateRequest
```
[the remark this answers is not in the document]
wabicai: 已在 d5f7e356f 同步三处维护文档：Stage 携带 targets，随后发送空 DeviceFirmwareUpdateRequest。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 已在 d5f7e356f 删除 host materializedEntries 与 Core 解包结果的严格集合相等校验。现在以已验证归档字节为权威：仍校验完整归档 SHA-256，并直接从该归档解析 manifest、校验选定文件内容；wrapper path 和额外 metadata 不再误报 mismatch。对应 wrapper + extra entry 回归测试已通过。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 当前实现不会让裸 Success 完成轮询。Success 仅设置 installEvidenceObserved 并继续等待；完成仍要求 requested-target 状态通过校验，或重启后正常运行态结合版本变化/安装证据。这里保留 returnAfterWrite 是为了兼容设备写入后立即重启或断链。

packages/hd-transport/src/types/messages.ts
wabicai: 已在 d5f7e356f 修复：低层 DeviceFirmwareUpdate 先通过 DeviceFirmwareUpdateStage 发送 targets，收到 Success 后再以 returnAfterWrite 发送空 DeviceFirmwareUpdateRequest。测试同时验证调用顺序与参数。

packages/connect-examples/expo-example/src/views/FirmwareScreen/firmwareUpdateMessages.ts
wabicai: 这是示例模块中的临时展示策略，当前产品决定保持 installingFirmware 映射到 99%；该 Expo 模块后续会整体删除，因此本 PR 不再增加设备/协议上下文与兼容分支。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 这里保留 current-install gate 是有意的：用户在设备上确认 Install 后才进入安装阶段，正常流程有足够时间观察 PENDING/IN_PROGRESS；直接接受首轮 FINISHED 无法与上一轮残留记录区分，会产生更危险的假成功。当前不放宽该门控。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 这个问题与上面的 legacy handshake 线程相同。当前 ProtocolInfo 没有可可靠区分两套复用消息 ID registry 的 capability/schema revision；按版本猜测会引入新的误判。当前 PR 按新固件 registry 实现，旧 loader 迁移需固件先提供明确能力位后单独处理，所以线程先保持开放。

packages/connect-examples/expo-example/src/views/FirmwareScreen/firmwareUpdateMessages.ts
wabicai: 这是 Expo 示例模块中的临时展示策略，当前决定保留 installingFirmware 统一显示 99%；该模块后续会删除，本 PR 不再扩展 Pro2 V4 上下文传递。

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -3148,65 +3128,12 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
-        }
-      } else {
-        throw error;
-      }
-    }
-    this.protocolV2LegacyDirectUpdate = false;
-    // A Success ACK or an install-time disconnect both end confirmation. In the
-    // latter case only status polling may establish the final outcome.
+    const commands = this.device.getCommands();
+    await commands.typedCall('DeviceFirmwareUpdateStage', 'Success', { targets });
```
originalix, line 3133: P1: [Preserve legacy Pro2 firmware update compatibility]

When connected to a loader using the previous Protocol V2 message IDs, `DeviceFirmwareUpdateStage` is encoded as ID 61000, which that firmware handles as the old side-effecting `DeviceFirmwareUpdateRequest`. The call can start installation and disconnect before returning `Success`; because this path awaits Stage without the former legacy interruption handling, the high-level update fails instead of entering reconnect and status polling even though device mutation has begun.

Keep a capability-based legacy branch that uses the old request and expected-disconnect handling, and only use Stage plus the empty trigger after the active firmware advertises the new handlers.

packages/hd-transport/src/protocols/v2/session.ts
```diff
@@ -271,6 +272,10 @@ export class ProtocolV2Session {
         logger?.error?.(`${logPrefix} write metrics callback failed: ${String(error)}`);
       }
 
+      if (callOptions.returnAfterWrite) {
```
originalix, line 275: P1: [Do not leave the fire-and-forget response in the active link]

When a `returnAfterWrite` request later produces its Protocol V2 ACK or terminal frame, this return has already released the serialized call without consuming that frame. The next status query or SDK command can then read the stale ACK under a different request sequence, causing an ACK-sequence failure and link invalidation, or consume the old terminal response in the wrong call.

Add a protocol-aware trigger completion mode that keeps ownership of the response stream until it consumes the matching ACK, then returns without waiting for the terminal business response.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -113,6 +113,13 @@ const PROTOCOL_V2_RESOURCE_MANIFEST_MAX_BYTES = 1024 * 1024;
 const PROTOCOL_V2_RESOURCE_FILE_MAX_COUNT = 512;
 const PROTOCOL_V2_RESOURCE_TOTAL_MAX_BYTES = 256 * 1024 * 1024;
 
+const getProtocolV2LocalResourceArchivePath = (entryName: string) => {
+  const match = entryName.match(
+    /(?:^|\/)((?:bundles\/|loaders\/(?:bootloader|rom)\/).+\.okpkg)$/iu
+  );
+  return match?.[1];
+};
```
sidmorizon, line 121: **Local firmware resource archives containing bootloader loader packages are rejected outright**

Severity: `non-severe`

Resource packages found under the bootloader loader folder are collected into the generated package list (`getProtocolV2LocalResourceArchivePath` at `packages/core/src/api/FirmwareUpdateV4.ts:116-121`) even though only one specific bootloader package name is ever accepted later, so the whole update aborts with a confusing manifest error instead of proceeding.
Impact: Users updating from a local resource archive that contains any bootloader loader package other than the single allowed one get a failed update with a misleading error.

<details>
<summary>Path allowlist mismatch between ZIP scanning and manifest validation</summary>

When no `manifest.json` is present, `prepareProtocolV2LocalResourceArchive` builds synthetic manifest entries with `device_path: vol0:/<archivePath>` for every entry matched by the new regex, which accepts `loaders/bootloader/<any>.okpkg` (`packages/core/src/api/FirmwareUpdateV4.ts:116-121`, `packages/core/src/api/FirmwareUpdateV4.ts:1309-1321`).

The synthesized manifest is later re-parsed in `prepareProtocolV2ResourceArchiveSources` via `parseProtocolV2ResourceManifest`, whose `isAllowedManifestDevicePath` only permits `vol0:/bundles/**`, `vol0:/loaders/rom/**`, and the exact path `vol0:/loaders/bootloader/boot_resource.okpkg` (`packages/core/src/protocols/protocol-v2/resources.ts:61-79`). Any other bootloader package therefore throws `Invalid Pro2 resource manifest files[i].device_path`, failing the entire update rather than skipping the unsupported file.

Either the collection regex should be narrowed to the supported bootloader package, or the allowlist extended to `vol0:/loaders/bootloader/`.
</details>

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 忽略，不用考虑兼容

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 忽略，不会存在这种情况。

packages/hd-transport/src/protocols/v2/session.ts
wabicai: 可以忽略，这里触发之后，他其实是卡住的状态。需要我们自己轮训

originalix (approved):

PR #871, opened 2026-08-12
fix: restore Classic BLE V1 handshake(OK-60130)

originalix:
## Summary
- Restore `Initialize` as the first Protocol V1 BLE probe for React Native and Electron transports.
- Keep probe timeout classification and diagnostics aligned with the restored handshake.
- Publish the fix as `1.2.0-alpha.111` with the npm `next` dist-tag.

## Intent & Context
OneKey Classic devices could be discovered and physically connected over BLE, but protocol detection failed on iOS, Android, and Desktop. The device log for OK-60130 showed the first V1 `GetFeatures` probe timing out after 1 second, followed by an unsuccessful V2 fallback and eventual protocol-detection failure.

This PR narrowly reverts the first-handshake change introduced with Protocol V2 support. It does not revert any other part of that work.

## Root Cause
The active V1/V2 detection flow changed the first V1 probe from `Initialize` to `GetFeatures`. Classic firmware did not respond to this `GetFeatures` probe within the probe timeout, so the transport treated the V1 attempt as failed and tried Protocol V2. The resulting fallback could not identify a Classic device and the connection failed.

## Design Decisions
- Revert only the initial V1 probe command in the React Native and Electron BLE transports.
- Update the adjacent timeout classification, detection error text, and diagnostic logs so an `Initialize` probe timeout retains the intended fallback behavior.
- Leave Core device-state reads, Protocol V2 framing/session behavior, USB transports, and all other PR #845 changes untouched.
- Preserve the existing 1-second probe timeout; the failure was tied to the probe command rather than a general timeout policy change.

The trade-off is that `Initialize` restores the legacy behavior that may reset an existing wallet session during protocol probing. This is intentional for Classic compatibility.

## Changes Detail
- `packages/hd-transport-react-native/src/index.ts`: restore `Initialize` for the V1 probe and align probe timeout/error handling.
- `packages/hd-transport-web-device/src/electron-ble-transport.ts`: apply the same handshake restoration for Desktop BLE.
- `packages/hd-transport-web-device/__tests__/electron-ble-transport.test.ts`: assert that repeated V1 acquisition sends message type `Initialize` (`0x0001`) and does not invoke the V2 writer.
- Package manifests: bump the SDK prerelease version to `1.2.0-alpha.111`.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (iOS and Android), Desktop
- **Risk Areas**: Initial BLE protocol detection for Protocol V1 devices; legacy wallet-session reset semantics of `Initialize`

## Test plan
- [x] `yarn agent:check --profile commit`
- [x] Electron BLE targeted suite: 16 tests passed
- [x] React Native BLE targeted suite: 39 tests passed
- [x] `yarn check-versions`
- [x] `yarn bootstrap`
- [x] `yarn build` (25 packages passed)
- [x] Publish workflow completed successfully: https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/31591036904
- [x] npm `next` resolves to `1.2.0-alpha.111` and the published package `gitHead` matches `7619f316471ad423fb83f68e82e1aec89f34ac0a`

Related issue: https://onekeyhq.atlassian.net/browse/OK-60130


packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -579,13 +579,11 @@ export default class ElectronBleTransport {
 
     try {
       this.deviceProtocol.set(uuid, 'V1');
-      // GetFeatures identifies Protocol V1 without resetting an existing wallet
-      // session before Core has a chance to restore a hidden wallet.
-      await this.callProtocolV1(uuid, 'GetFeatures', {}, { timeoutMs: PROTOCOL_PROBE_TIMEOUT_MS });
+      await this.callProtocolV1(uuid, 'Initialize', {}, { timeoutMs: PROTOCOL_PROBE_TIMEOUT_MS });
```
originalix: P2: [Electron BLE retains the one-second Initialize probe timeout]

When a Classic device takes more than one second to answer this `Initialize`, the existing Electron timeout treats the valid V1 probe as failed. Detection then reconnects and tries Protocol V2, so acquisition can still fail for a supported V1 device. The React Native path in this PR raises the equivalent budget to three seconds, but Electron keeps the old one-second value.

Give the Electron V1 `Initialize` probe the same validated three-second budget and add a delayed-response test that answers after one second but before the new deadline.

wabicai (approved):

PR #878, opened 2026-08-14
fix: serialize WebUSB bootloader polling (OK-60205)

originalix:
## What changed

- serialize `FirmwareUpdateV2` bootloader probes across WebUSB and BLE polling ticks
- keep polling paused while discovery/acquire is still in flight, then resume after it settles
- add a WebUSB regression test that proves only one probe can be active at a time
- prepare workspace package versions as `@onekeyfe/*@1.2.0-alpha.123` for publication from `onekey` after merge

## Why

During a Universal-to-Bitcoin-Only firmware switch on Classic-family devices, the polling interval could start another WebUSB probe before the previous probe completed. The overlapping transport operations raced the same connection and surfaced as a misleading Protocol V1 mismatch.

## Impact

Firmware switching no longer overlaps bootloader probes, while failed probes still release the guard so later polling ticks can retry.

## Validation

- `yarn agent:check --profile pr`
- focused bootloader polling Jest regression
- `yarn check-versions` confirms all 30 packages are `1.2.0-alpha.123`
- npm registry confirmed `1.2.0-alpha.123` was unused before the bump
- `1.2.0-alpha.123` is intentionally not published or tagged in this PR; published `1.2.0-alpha.122` was used for local App verification


packages/core/src/api/FirmwareUpdateV2.ts
```diff
@@ -276,10 +276,9 @@ export default class FirmwareUpdateV2 extends BaseMethod<Params> {
     this.checkPromise = createDeferred();
     const env = DataManager.getSettings('env');
     const isBleReconnect = connectId && DataManager.isBleConnect(env);
-    // Tracks the in-flight BLE probe so interval ticks never overlap: a superseded
-    // Initialize left pending on the V1 transport is exactly what later times out
-    // and used to tear down the connection mid-firmware-upload.
-    let bleProbeInFlight = false;
+    // Bootloader probes can outlive the polling interval. Keep discovery and
```
originalix, line 279: P2: [The serialization fix leaves other firmware pollers unprotected]

When `firmwareUpdateV3` or legacy `firmwareUpdate` uses WebUSB and an `enumerate()`/`DevicePool.getDevices()` probe lasts longer than the polling interval, their unchanged callbacks can still start concurrent probes on the same connector. `FirmwareUpdateBaseMethod` currently guards only its BLE branch, and `FirmwareUpdate` has another unguarded non-BLE poller, so those public paths can still race device cache and connection state and fail the bootloader handoff or upload.

Please apply this generic in-flight guard to `FirmwareUpdateBaseMethod` and the legacy `FirmwareUpdate` poller, then include those methods in the serialization regression test matrix.

originalix:

@originalix

Reviewed commit `27a4b3f8cf09`. This report always reflects the latest reviewed changes.

### Review summary

The PR changes Core's Protocol V1 reboot-to-bootloader orchestration so FirmwareUpdateV2 serializes WebUSB discovery and BLE acquire probes, adds a focused polling regression test, and synchronizes monorepo package versions at 1.2.0-alpha.123. Equivalent WebUSB polling paths in the shared FirmwareUpdateV3 base and legacy FirmwareUpdate remain unprotected.

**What needs attention:** Apply the serialization rule consistently to every public Protocol V1 firmware updater and extend the regression matrix before merging.

### Issues to address

- **P2 · The serialization fix leaves other firmware pollers unprotected**
  - **Impact:** When FirmwareUpdateV3 or legacy FirmwareUpdate polls through WebUSB and an enumerate/getDevices probe lasts longer than the two-second interval, their unchanged callbacks can start another probe on the same connector. The probes can race DevicePool and transport connection state, causing the reboot-to-bootloader handoff or subsequent firmware upload to fail.
  - **Suggested action:** Apply the generic in-flight guard to FirmwareUpdateBaseMethod and the legacy FirmwareUpdate poller, then include both public methods in the WebUSB serialization regression matrix.

### Validation gaps

- `yarn check-versions` passed for all 30 versioned packages at `1.2.0-alpha.123`.
- The diff adds no outbound request, host, sensitive logging, protobuf, or framing changes.
- The focused Jest test and Core lint were not executed because this workspace does not contain the installed Jest and ESLint binaries.
- No physical-device WebUSB reboot and upload handoff was validated.
- The added serialization test covers only FirmwareUpdateV2, not FirmwareUpdateV3 or legacy FirmwareUpdate.

wabicai (approved):

packages/core/src/api/FirmwareUpdateV2.ts
originalix, line 279: P2: [The serialization fix leaves other firmware pollers unprotected]

When `firmwareUpdateV3` or legacy `firmwareUpdate` uses WebUSB and an `enumerate()`/`DevicePool.getDevices()` probe lasts longer than the two-second interval, their unchanged callbacks can start concurrent probes on the same connector. `FirmwareUpdateBaseMethod` guards only its BLE branch, and `FirmwareUpdate` has another unguarded non-BLE poller, so these paths can still race device-cache and connection state and fail the bootloader handoff or subsequent upload.

Please apply this generic in-flight guard to `FirmwareUpdateBaseMethod` and the legacy `FirmwareUpdate` poller, then include both methods in the WebUSB serialization regression matrix.

PR #879, opened 2026-08-15
fix: harden Protocol V2 firmware updates and release resources

wabicai:
## Summary

- harden the Protocol V2 firmware installation flow across confirmation, reconnect, polling, and transport-session boundaries
- bind Pro2 and Neo resource archives to the selected `firmware-v1` release instead of reading a device-global resource entry
- validate release resource metadata and prepared resource archive readers before downloading or reading an approved ZIP artifact
- align USB, WebUSB, React Native, low-level, and Electron BLE transports with the reconnect-safe Protocol V2 session lifecycle
- document the relevant Protocol V2 device-management and firmware UI events

## Root cause

Protocol V2 installation spans multiple device modes and transport reconnects. The previous flow could lose confirmation or session state while the device rebooted, and polling could confuse stale status records with the current installation. Resource metadata was also modeled globally per device, so an archive was not guaranteed to match the firmware release being installed. Separately, the prepared-resource path allowed an optional reader to reach an API that requires a concrete reader.

### Firmware installation and progress

- make the install request return after the write so the SDK can continue tracking the device through the expected disconnect
- distinguish current installation evidence from stale or unrelated status records
- preserve Protocol V2 session/link state across supported transport reconnects
- emit the confirmation and firmware progress events needed by clients to represent the installation lifecycle correctly

### Release-scoped resources

- parse and retain `resources` on each Pro2/Neo `firmware-v1` release
- resolve the resource archive from the selected firmware release and firmware type
- validate the archive URL, SHA-256 metadata, prepared artifact, and resource reader before installation
- ignore invalid optional resource metadata without breaking base device communication, while surfacing an explicit error when that resource is required

### Transport and documentation updates

- make Protocol V2 reconnect behavior consistent across USB, WebUSB, React Native, low-level, and Electron BLE transports
- add focused transport/session regression coverage
- update Protocol V1/V2, Pro2 device-management, and SDK event documentation

## Impact

Pro2 and Neo updates now install resources that belong to the exact firmware release being applied, maintain the expected Protocol V2 lifecycle through reconnects, and provide deterministic confirmation/progress behavior to App clients. Invalid or unavailable resource artifacts fail with a targeted error instead of producing an ambiguous runtime failure.

## Validation

- `yarn test firmware-prepared-resource-archive.test.ts --runInBand` — 6 tests passed
- added focused coverage for Protocol V2 resources, install polling, link management, sessions, and Electron BLE reconnect behavior
- focused ESLint checks passed
- `@onekeyfe/hd-core` build passed without the `FirmwarePreparedResourceArchive` TS2322 error

### DIFF
--- a/packages/core/src/api/FirmwareUpdateV4.ts
+++ b/packages/core/src/api/FirmwareUpdateV4.ts
@@ -915,7 +915,8 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
       }
       if (wantsResources) {
         this.params.resourceArchiveBinary = await this.downloadRemoteProtocolV2ResourceArchive(
-          deviceFeatures
+          deviceFeatures,
+          firmwareType
         );
         resourceMemoryHost = await this.prepareProtocolV2LocalMemoryHost({
           features: deviceFeatures,
@@ -1880,7 +1881,10 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     };
   }
 
-  private async downloadRemoteProtocolV2ResourceArchive(features: Features): Promise<ArrayBuffer> {
+  private async downloadRemoteProtocolV2ResourceArchive(
+    features: Features,
+    firmwareType: EFirmwareType
+  ): Promise<ArrayBuffer> {
     const deviceType = getDeviceType(features);
     if (deviceType !== EDeviceType.Pro2 && deviceType !== EDeviceType.Neo) {
       throw ERRORS.TypedError(
@@ -1888,7 +1892,8 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         'Protocol V2 resource archive requires a Pro2 or Neo device'
       );
     }
-    const source = DataManager.getProtocolV2ResourceSource(deviceType);
+    const release = DataManager.getFirmwareLatestRelease(features, firmwareType);
+    const source = DataManager.getProtocolV2ResourceSource(release);
     const expectedSha256 = normalizeProtocolV2Hex(source?.archiveSha256);
     if (
       !source?.archiveUrl ||
--- a/packages/core/src/api/FirmwareUpdateV4.ts
+++ b/packages/core/src/api/FirmwareUpdateV4.ts
@@ -9,7 +9,7 @@ import {
 } from '@onekeyfe/hd-transport';
 import { sha256 } from '@noble/hashes/sha256';
 
-import { FirmwareUpdateTipMessage, UI_REQUEST } from '../events/ui-request';
+import { FirmwareUpdateTipMessage, UI_REQUEST, createUiMessage } from '../events/ui-request';
 import { validateParams } from './helpers/paramsValidator';
 import {
   LoggerNames,
@@ -71,6 +71,7 @@ import type {
   IProtocolV2FirmwareComponent,
   IProtocolV2ResourceManifestFile,
   IVersionArray,
+  KnownDevice,
 } from '../types';
 import type { FirmwareByteSource } from './firmware/FirmwareArtifactSource';
 import type {
@@ -2375,9 +2376,13 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
       target_id: item.targetId,
       path: item.path,
     }));
-    await this.protocolV2StartFirmwareUpdate({ targets });
-    await wait(PROTOCOL_V2_INSTALL_STATUS_INITIAL_DELAY);
-    await this.waitForProtocolV2FirmwareUpdateComplete(targets, true);
+    try {
+      await this.protocolV2StartFirmwareUpdate({ targets });
+      await wait(PROTOCOL_V2_INSTALL_STATUS_INITIAL_DELAY);
+      await this.waitForProtocolV2FirmwareUpdateComplete(targets, true);
+    } finally {
+      this.device?.clearCancelableAction();
+    }
   }
 
   private async protocolV2SourceUpdateProcess({
@@ -2637,7 +2642,9 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     const expectedPaths = new Map(targets.map(target => [target.target_id, target.path]));
     const startTime = Date.now();
     let lastError: unknown;
-    let shouldReconnect = true;
+    // DeviceFirmwareUpdateRequest leaves the current loader link usable for status polling.
+    // Reconnecting here probes DeviceInfo, which loaders reject while installation is active.
+    let shouldReconnect = false;
     let deviceInfo: ProtocolV2DeviceInfo | undefined;
     let missingTargetStatusSince: number | undefined;
     let missingTargetStatusKey: string | undefined;
@@ -2684,16 +2691,15 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
             statusResponse.type === 'DeviceFirmwareUpdateStatus'
               ? ((statusResponse.message.records ?? []) as ProtocolV2FirmwareUpdateStatusTarget[])
               : [];
-          if (
-            statusTargets.some(target => {
-              const targetId = normalizeProtocolV2TargetId(target.target_id);
-              return (
-                targetId !== undefined &&
-                expectedTargetIds.has(targetId) &&
-                isProtocolV2TargetStatusInProgress(target.status)
-              );
-            })
-          ) {
+          const hasInProgressTarget = statusTargets.some(target => {
+            const targetId = normalizeProtocolV2TargetId(target.target_id);
+            return (
+              targetId !== undefined &&
+              expectedTargetIds.has(targetId) &&
+              isProtocolV2TargetStatusInProgress(target.status)
+            );
+          });
+          if (hasInProgressTarget) {
             currentInstallStatusObserved = true;
           }
           const hasMatchingTargetStatus = statusTargets.some(target => {
@@ -3138,9 +3144,20 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     this.protocolV2LastRuntimeProbeFeatures = undefined;
     const commands = this.device.getCommands();
     await commands.typedCall('DeviceFirmwareUpdateStage', 'Success', { targets });
+    this.device.setCancelableAction(() => commands.cancelDevice());
+    const interaction = this.device.createProtocolV2UiPhaseMetadata('button', 'start');
+    this.postMessage(
+      createUiMessage(UI_REQUEST.REQUEST_BUTTON, {
+        device: this.device.toMessageObject() as KnownDevice,
+        source: 'method-lifecycle',
+        reason: 'firmware-update',
+        completion: 'operation-completed',
+        deviceOnly: true,
+        operation: this.name,
+        ...(interaction ? { interaction } : {}),
+      })
+    );
     await commands.call('DeviceFirmwareUpdateRequest', {}, { returnAfterWrite: true });
-    this.postTipMessage(FirmwareUpdateTipMessage.FirmwareUpdating);
-    this.postProgressMessage(0, 'installingFirmware');
   }
 
   private async protocolV2Reboot(rebootType: DeviceRebootType) {
--- a/packages/hd-transport-web-device/src/electron-ble-transport.ts
+++ b/packages/hd-transport-web-device/src/electron-ble-transport.ts
@@ -493,7 +493,8 @@ export default class ElectronBleTransport {
     for (let i = 0; i < probeOrder.length; i += 1) {
       const protocol = probeOrder[i];
       if (i > 0) {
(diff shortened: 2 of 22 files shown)


packages/hd-transport/src/protocols/v2/session.ts
```diff
@@ -215,6 +219,47 @@ export class ProtocolV2Session {
     return result;
   }
 
+  sendFlowControl(name: string, data: Record<string, unknown>): Promise<MessageFromOneKey> {
```
originalix, line 222: P1: [Flow-control replies can satisfy the wrong Protocol V2 call]

When a user cancels a pending Protocol V2 prompt, `sendFlowControl()` only writes `Cancel` and returns `WriteCompleted`. The original call is still reading from the same session, and `executeCall()` later accepts any terminal frame of the expected type without matching `decoded.seq` back to the request sequence.

If firmware emits `Success` for the `Cancel`, that `Success` can be consumed as the still-running business call's completion. The cancellation path then reports the wrong outcome and leaves the session out of sync for whatever reads next.

Please either drain or route flow-control replies separately, or enforce request/response sequence matching before a terminal frame is returned to the active call.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -914,7 +915,8 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
       }
       if (wantsResources) {
         this.params.resourceArchiveBinary = await this.downloadRemoteProtocolV2ResourceArchive(
```
originalix, line 917: P2: [Local firmware artifacts can be paired with the wrong resource archive]

In this branch the update keeps the caller-supplied local install items, but the resource ZIP is still fetched remotely. After this PR moved resources onto individual releases, `downloadRemoteProtocolV2ResourceArchive()` now resolves the ZIP from `DataManager.getFirmwareLatestRelease(features, firmwareType)` instead of from the version represented by those local binaries.

That means a local firmware bundle can now be combined with the latest remote resource archive even when they belong to different releases. The result is a mixed firmware/resource set that can fail installation or stage the wrong assets.

Please either reject this mixed mode unless a prepared plan binds both sides, or derive the resource release from the selected local artifact versions and verify it matches before downloading.

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -540,36 +541,13 @@ export default class ElectronBleTransport {
...
-      this.connectedDevices.add(uuid);
-      await window.desktopApi?.nobleBle?.subscribe(uuid);
-      await this.refreshBlePacketCapacity(uuid);
-    } catch (error) {
-      this.Log?.debug(`[Electron BLE] reconnect after Protocol ${protocol} probe failed:`, error);
-      throw error;
+    // A timed-out V1 probe retires its renderer notification token so a late V1
+    // response cannot satisfy the V2 probe. Install a fresh listener without
+    // touching the native GATT subscription or physical connection.
+    if (!this.notificationCleanups.has(uuid)) {
```
originalix, line 547: P2: [Electron BLE fallback keeps the stale notification token alive]

After a timed-out V1 probe, this code intends to rotate the renderer notification token so late V1 packets cannot satisfy the V2 probe. In the normal acquire path `notificationCleanups` already has an entry, so this branch never calls `createNotificationSubscription()` and the old token stays active.

A late V1 notification can therefore still be delivered after `deviceProtocol` flips to `V2`, where it is parsed as a V2 frame and can poison or time out the fallback probe. That makes the new no-reconnect pairing flow flaky on the exact onboarding path it is trying to fix.

Please bump the notification token on every probe switch, or reinstall the renderer callback unconditionally while keeping the native GATT subscription intact.

originalix:

@wabicai

Reviewed commit `14e5b24b13a4`. This report always reflects the latest reviewed changes.

### Review summary

This PR makes Protocol V2 resource metadata release-scoped, adds Protocol V2 flow-control `post()` support across transports so firmware-update confirmation can be canceled without waiting on the original call, and changes Electron BLE V1->V2 probe fallback to reuse the existing GATT link instead of reconnecting. The docs and tests were updated to describe the new firmware-update UI event and the revised probe/install behavior.

**What needs attention:** Re-check the new Protocol V2 cancellation path, release-to-resource binding, and Electron BLE fallback probe race before merge. Those are the places where the diff changes state-machine behavior rather than just wiring.

### Issues to address

- **P1 · Flow-control replies can satisfy the wrong Protocol V2 call**
  - **Impact:** The new `sendFlowControl()` path writes `Cancel` without owning any response handling, while the active `ProtocolV2Session.executeCall()` still accepts any terminal frame of the expected type without matching it back to the originating request sequence. If firmware emits `Success` for `Cancel`, the waiting business call can complete as if its own request succeeded, which corrupts cancellation behavior in firmware update and other interactive Protocol V2 flows.
  - **Suggested action:** Route flow-control responses away from the active call or require terminal-response sequence matching before returning from `executeCall()`, then add a regression test where a `Cancel` reply arrives before the original interactive call resolves.
- **P2 · Local firmware artifacts can be paired with the wrong resource archive**
  - **Impact:** After resources became release-scoped, `FirmwareUpdateV4` still downloads the remote resource ZIP from `DataManager.getFirmwareLatestRelease(features, firmwareType)` even when the firmware components come from caller-supplied local binaries. That can mix one release's staged firmware with another release's resource bundle, producing mismatched update content on a public firmware-update path.
  - **Suggested action:** Either reject mixed local-firmware plus remote-resource updates unless a prepared plan binds both, or derive the resource release from the selected local artifact versions and verify it matches before downloading.
- **P2 · Electron BLE fallback keeps the stale notification token alive**
  - **Impact:** The fallback path claims to rotate the renderer notification token after a timed-out V1 probe, but it only creates a new subscription when no cleanup entry exists. In the normal acquire path a cleanup already exists, so late V1 packets still pass the old token check and can be fed into the V2 probe, making the new no-reconnect pairing flow flaky.
  - **Suggested action:** Rotate the notification token on every probe switch, either by reinstalling the renderer callback unconditionally or by bumping the token independently of subscription cleanup, and add a late-V1-after-timeout regression test.

### Validation gaps

- No regression test shows what happens if a Protocol V2 `Cancel` reply arrives while another call is waiting for `Success`.
- No test covers a late V1 BLE notification arriving after V1 probe timeout while the same GATT subscription is reused for the V2 probe.
- No test covers the public mixed mode where local Protocol V2 component binaries are combined with a remotely fetched resource archive after resources became release-scoped.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/hd-transport/src/protocols/v2/session.ts
wabicai: 误报。不可能存在这种情况

packages/hd-transport-web-device/src/electron-ble-transport.ts
wabicai: 误报。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 忽略，真正的产品和实际场景不会有这种情况。

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -540,36 +541,13 @@ export default class ElectronBleTransport {
...
+      const cleanup = this.createNotificationSubscription(uuid);
+      this.notificationCleanups.set(uuid, cleanup);
```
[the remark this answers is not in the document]
wabicai: 忽略。不可能存在这种情况

packages/hd-transport/src/protocols/v2/session.ts
```diff
@@ -215,6 +219,47 @@ export class ProtocolV2Session {
...
+    if (maxFrameBytes !== undefined && frame.length > maxFrameBytes) {
+      return Promise.reject(
+        new Error(`Protocol V2 frame too large for transport: ${frame.length} > ${maxFrameBytes}`)
+      );
+    }
+
+    return this.serializeWrite(() => writeFrame(frame, context)).then(() => ({
+      type: 'WriteCompleted',
+      message: {},
+    }));
```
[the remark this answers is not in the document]
wabicai: sendFlowControl 可以不用处理。这里目前没有看到能复现的场景。属于异常情况。

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2637,7 +2647,9 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     const expectedPaths = new Map(targets.map(target => [target.target_id, target.path]));
     const startTime = Date.now();
     let lastError: unknown;
-    let shouldReconnect = true;
+    // DeviceFirmwareUpdateRequest leaves the current loader link usable for status polling.
+    // Reconnecting here probes DeviceInfo, which loaders reject while installation is active.
+    let shouldReconnect = false;
```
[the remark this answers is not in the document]
wabicai: BLE ACK 会在首次 polling 前的一秒窗口内到达，并由 prepareCall() reset；terminal Success 则是 polling 明确支持的响应。

originalix (approved):

packages/hd-transport/src/protocols/v2/session.ts
```diff
@@ -215,6 +219,47 @@ export class ProtocolV2Session {
...
+  }
```
sidmorizon, line 255: **Cancelling a firmware confirmation can wait forever when the connection stalls**

Severity: `non-severe`

The cancel message is sent with no time limit at all (`sendFlowControl` at `packages/hd-transport/src/protocols/v2/session.ts:251-255`), unlike every other device message, so a stalled connection leaves the cancel attempt waiting indefinitely.
Impact: When a user cancels the on-device firmware confirmation and the link is wedged, the cancel never finishes and the app stays stuck instead of failing.

<details>
<summary>Missing watchdog around the write-only flow-control frame</summary>

`ProtocolV2Session.call()` wraps its whole request/response flow in `withProtocolTimeout(...)` with `PROTOCOL_V2_DEFAULT_RESPONSE_TIMEOUT_MS` and an `AbortSignal` that is actually aborted on timeout (`packages/hd-transport/src/protocols/v2/session.ts:236-240`, 370+). The new `sendFlowControl` builds a fresh `AbortController` that is never aborted, passes it into `writeFrame`, and returns `this.serializeWrite(() => writeFrame(frame, context))` without any timeout wrapper.

The path is reachable from user cancellation: `Device.cancel()` awaits `cancelableAction` (`packages/core/src/device/Device.ts:1525-1534`), which `FirmwareUpdateV4.protocolV2StartFirmwareUpdate` now sets to `commands.cancelDevice()` (`packages/core/src/api/FirmwareUpdateV4.ts:3152`). `cancelDevice()` reaches `cancelDeviceInPrompt(device, false)`, which calls `transport.post(session, 'Cancel', {})` (`packages/core/src/device/DeviceCommands.ts:138`), and the new `post()` implementations in the BLE/USB transports route V2 to `sendFlowControl`. A BLE/USB write that never settles therefore blocks the cancel path with no bound, whereas the previous `post()` → `call()` path was bounded by the shared 5-minute call watchdog.

</details>

packages/hd-transport/src/protocols/v2/link-manager.ts
```diff
@@ -78,6 +78,17 @@ export class ProtocolV2LinkManager<Key> {
...
+  sendFlowControl(
+    key: Key,
+    createAdapter: () => ProtocolV2LinkAdapter,
+    name: string,
+    data: Record<string, unknown>
+  ): Promise<MessageFromOneKey> {
+    const generation = this.generations.get(key) ?? 0;
+    this.assertCallGeneration(key, generation);
+    return this.getOrCreateLink(key, createAdapter).session.sendFlowControl(name, data);
+  }
```
sidmorizon, line 90: **A failed cancel send leaves a broken device connection marked as usable**

Severity: `non-severe`

A send failure for the cancel message is passed straight back to the caller without marking the connection as broken (`sendFlowControl` at `packages/hd-transport/src/protocols/v2/link-manager.ts:81-90`), unlike every other message send, so later commands keep reusing a dead connection.
Impact: After a cancel fails because the device link died, the following device operations fail with confusing errors until the user manually reconnects.

<details>
<summary>Flow-control path skips the link-fatal classification used by executeCall</summary>

`ProtocolV2LinkManager.call()` funnels through `executeCall`, which classifies errors via `options.classifyError` and invalidates the link (resetting the adapter, assembler, native connection and protocol caches) when the error is link-fatal. The new `sendFlowControl` only checks the current generation and then calls `session.sendFlowControl(...)` directly, so an I/O/disconnect/generation failure raised inside `writeFrame` never triggers `invalidateLink`. The link object stays in `this.links`, and the next `call()` reuses the same adapter/session over a connection that the platform layer already considers dead.

This conflicts with the repository rule in `AGENTS.md` ("Treat timeout, disconnect, I/O, generation, framing, CRC, and sequence failures as link-fatal") and with `docs/protocol/protocol-v1-v2.md`, which requires invalidating the Link before cleaning up reads, assembler and platform connection.

</details>

PR #881, opened 2026-08-16
fix: refresh Pro2 state and serialize BLE cancellation

wabicai:
## What changed

- Refresh Protocol V2 status and settings after every supported settings mutation, including wallpaper changes.
- Reconcile stale label and passphrase state through the canonical `DeviceState` event.
- Send an explicit Protocol V2 BLE `Cancel` when the App closes an active operation without a prompt callback.
- Wait for cancellation cleanup and serialize per-device BLE acquire/release/disconnect lifecycle operations.
- Cold-reconnect and retry when protocol detection state is missing after repeated BLE interactions.

## Why

Pro2 settings could diverge between the device and App until a later explicit refresh. Rapid BLE interaction, retry, and cancellation could also overlap teardown with reacquisition, leaving the transport stuck or making App-side Cancel ineffective.

## Impact

Pro2 settings mutations now publish a current device snapshot, while BLE cancellation fully tears down the active operation before the same device is reused.

## Related App change

- https://github.com/OneKeyHQ/app-monorepo/pull/12899

## Validation

- Core focused tests: 319 passed, 4 skipped.
- React Native BLE transport tests: 40 passed.
- ESLint passed for all changed files.
- Local SDK build was consumed by the App and validated on an iPhone 16 Pro running iOS 26.6.


wabicai:
Published `1.2.0-alpha.130` to npm with the `next` dist-tag via GitHub Actions: https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/31931672626

sidmorizon:
@codex review

packages/core/src/core/index.ts
```diff
@@ -311,9 +304,9 @@ const waitForPendingPromise = async (
   if (pendingPromise) {
     Log.debug('pre pending call promise before call method, wait for it');
     try {
-      await waitWithTimeout(pendingPromise, 5 * 1000);
+      await pendingPromise;
```
originalix: P1: [Global pending-cancel waits can now deadlock the whole SDK instance]

`prePendingCallPromise` is a single Core-wide gate, and every later request waits here before it can start. This change removed the previous 5s fallback and now waits for that promise to settle unconditionally.

The problem is that the new `device.interruptionFromUser()` path also waits for full cleanup, while the BLE release/disconnect path still contains unbounded native awaits. If one cancel teardown wedges, this promise never resolves, `prePendingCallPromise` never clears, and every later call in the same SDK instance can block forever, including unrelated devices.

Please keep the stronger cleanup ordering, but restore a bounded escape hatch here or scope the pending-cancel gate per `connectId` so a stuck teardown cannot freeze the whole SDK.

sidmorizon:
@codex security review

originalix:

@wabicai

Reviewed commit `b6d3307b47c5`. This report always reflects the latest reviewed changes.

### Review summary

This PR hardens Protocol V2 BLE cancellation and reconnect behavior by tracking per-device pending cleanup in Core, serializing React Native BLE lifecycle operations, adding bounded native teardown with manager-reset fallback, and refreshing Protocol V2 settings state after on-device mutations. It also updates docs/tests around the new state flow and bumps published package versions to 1.2.0-alpha.133.

**What needs attention:** Re-check the Protocol V2 passphrase-setting success criteria and the cross-device effects of React Native BLE manager resets, especially for in-flight calls that were not part of the timed-out teardown.

### Issues to address

- **P1 · Passphrase settings can report success without changing the device**
  - **Impact:** The Protocol V2 passphrase branch now ignores the post-page read-back result. If the device closes the on-device page without actually changing the toggle, callers still receive success for a security-sensitive setting and may persist or act on a state the hardware never adopted.
  - **Suggested action:** Restore a post-refresh mismatch check for authoritative reads, while keeping the special handling only for the locked-after-disable case where the state cannot be verified.
- **P2 · BLE manager reset leaves active Protocol V1 calls unresolved**
  - **Impact:** A native teardown timeout now resets the shared React Native BLE manager for every cached session, but the reset path never rejects the transport's active Protocol V1 deferred. An unrelated in-flight call can therefore stall long after its underlying manager and subscriptions were destroyed.
  - **Suggested action:** When resetting the shared BLE manager, explicitly reject the active run promise or route the reset through a disconnect path that settles in-flight Protocol V1 calls before destroying the manager.

### Validation gaps

1. Targeted Jest suites were not runnable in this workspace because neither `yarn jest` nor `./node_modules/.bin/jest` was available.
2. I reviewed only the changed files plus directly related state-mapping behavior; there was no device-side verification on real BLE hardware.

originalix (dismissed):

packages/core/src/core/index.ts
```diff
@@ -311,9 +304,9 @@ const waitForPendingPromise = async (
...
     } catch (error) {
-      // ignore timeout error
+      // Cancellation is best-effort, but its teardown must settle before reuse.
     }
```
sidmorizon, line 328: 🟠 P1: A cancellation that never finishes cleaning up blocks every later device request forever

Severity: `severe`

The wait for a previous cancellation to finish is left with no time limit (`await pendingPromise` at `packages/core/src/core/index.ts:307`) even though that cleanup now also waits for the whole interrupted operation to unwind, so a cancellation that never settles blocks every subsequent request permanently.
Impact: If the device stops answering while the user cancels, the app stops responding to all further hardware requests until the SDK is torn down and re-created.

<details>
<summary>Why the unbounded wait can never be released</summary>

The promise stored by `setPrePendingCallPromise(device?.interruptionFromUser())` (`packages/core/src/core/index.ts:1184`) is awaited by every subsequent call before it can proceed (`packages/core/src/core/index.ts:351` and `:450`). Previously `waitWithTimeout(pendingPromise, 5 * 1000)` guaranteed forward progress; the PR removes that bound.

At the same time `Device.interruptionFromUser()` now additionally awaits `runCleanupPromise` (`packages/core/src/device/Device.ts:1550`), which only resolves once `_runInner` has fully unwound, i.e. after the in-flight `fn()` settles and after `this.release()` completes. The in-flight call itself has no built-in deadline: `ProtocolV2LinkManager` has no call timeout, and `DeviceCommands.call` only passes a timeout when a caller supplies one. The new BLE fallback `commands.cancelDevice()` is a fire-and-forget `post(session, 'Cancel', {})` (`packages/core/src/device/DeviceCommands.ts:243-258`), so if the device never answers, `fn()` never settles, `runCleanupPromise` never resolves, and `prePendingCallPromise` never settles. `commands.cancel()` is bounded at 10s, but that bound no longer covers the wait chain. The only recovery is `Core.dispose()`, which clears `prePendingCallPromise` (`packages/core/src/core/index.ts:1672`).

The same transport layer explicitly documents that native BLE teardown can stop settling (see the comment on `tearDownWedgedLink` in `packages/hd-transport-react-native/src/index.ts:1840-1846`), so an unsettled teardown is a realistic state.

</details>

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1632,6 +1649,26 @@ export default class ReactNativeBleTransport {
...
+  private async runLifecycleOperation<T>(uuid: string, operation: () => Promise<T>): Promise<T> {
+    const previousOperation = this.lifecycleOperations.get(uuid) ?? Promise.resolve();
+    let completeOperation!: () => void;
+    const operationGate = new Promise<void>(resolve => {
+      completeOperation = resolve;
+    });
+    const operationTail = previousOperation.catch(() => undefined).then(() => operationGate);
+    this.lifecycleOperations.set(uuid, operationTail);
+
+    await previousOperation.catch(() => undefined);
```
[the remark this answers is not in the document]
wabicai: 已在 `1274e5321` 处理：React Native BLE native teardown 统一设置 3 秒上限；超时后 reset `BleManager`，释放 lifecycle queue，并通过 generation guard 防止迟到的旧 cleanup 破坏新 link。新增 stalled native teardown 回归测试，验证后续 acquire/call 可继续。

packages/core/src/core/index.ts
wabicai: 已在 `1274e5321` 处理：pending cancellation gate 改为按 `connectId` 隔离，并保留 15 秒 bounded fallback；超时或完成时只移除仍匹配的 promise，避免旧 cleanup 清掉新任务，也不会阻塞其他设备。RN BLE cancel 超时还会主动 disconnect 触发 transport recovery。相关 5 个 suites 共 401 tests passed。

wabicai:
Review follow-up 已推送：`1274e5321` 增加 per-device bounded cancellation recovery，`81b472e98` 将全部 packages 升级到 `1.2.0-alpha.131`。聚焦回归结果：5 suites passed，401 tests passed。npm 已通过 GitHub Actions 以 `next` dist-tag 发布成功：https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/31947546459

originalix (commented):
@wabicai

Reviewed commit `81b472e983e8`.

- **P1 · Passphrase disable reports success without verifying the new state**
- **P2 · Per-device teardown timeout resets every BLE session**
- **P3 · hwk-demo still pins the pre-bump HWK package line**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/881#issuecomment-5306342042)

originalix (commented):
@wabicai

Reviewed commit `a0b0b1358f1b`.

- **P2 · Single-device teardown timeout destroys the shared RN BLE manager**
- **P2 · Wallpaper upload now fails on a best-effort settings refresh**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/881#issuecomment-5306342042)

packages/core/src/api/protocol-v2/DeviceUploadWallpaper.ts
```diff
@@ -143,6 +143,7 @@ export default class DeviceUploadWallpaper extends BaseMethod<DeviceUploadWallpa
     const response = await this.device.commands.typedCall('DeviceSettingsSet', 'Success', {
       settings: { wallpaper_path: this.path },
     });
+    await this.device.refreshProtocolV2SettingsAfterMutation();
```
originalix: P2: [Wallpaper upload now fails after the device already applied the wallpaper]

Once `DeviceSettingsSet` succeeds, the device has already accepted the new wallpaper path. This new `refreshProtocolV2SettingsAfterMutation()` call performs extra `DeviceStatusGet` / `DeviceSettingsGet` reads and now propagates any readback failure as the method result.

That turns a transient refresh problem into a false upload failure, so callers can show the wrong result or retry a large wallpaper transfer even though the device state already changed.

Please make this refresh best-effort, or separate cache reconciliation from the mutation result so a successful `DeviceSettingsSet` still returns success.

packages/core/src/api/device/DeviceSettings.ts
```diff
@@ -201,20 +200,7 @@ export default class DeviceSettings extends BaseMethod<ApplySettings> {
...
-          if (
-            updated.status.passphraseProtection !== requestedPassphrase &&
-            !lockedAfterDisabling
-          ) {
-            throw TypedError(
-              HardwareErrorCode.RuntimeError,
-              'Protocol V2 passphrase setting did not reach the requested value.'
-            );
-          }
+          await refreshStatusAndSettings();
```
originalix, line 203: P2: [Protocol V2 passphrase settings now return success even when readback still disagrees]

After `DeviceSettingsPageShow` returns, this branch no longer checks whether the refreshed status actually reached `requestedPassphrase`. When the device stays unlocked and the follow-up `DeviceStatusGet` still reports the old value, the method now returns `Success` anyway.

That turns a confirmed post-write mismatch into a false positive. Callers cannot tell whether the passphrase toggle really applied, and later wallet-session or unlock flows can run against stale security state.

Please keep the stale-state protection, but restore an explicit mismatch check for the cases where the device stays unlocked and the refreshed status still disagrees with the requested value.

packages/core/src/api/protocol-v2/DeviceUploadWallpaper.ts
wabicai: Handled in `4c0ded08a`. Once `DeviceSettingsSet` applies the wallpaper path, the post-apply status/settings refresh is now best-effort, so a transient read-back failure cannot report a false upload failure or trigger another large transfer. Added a focused regression test for successful apply followed by failed read-back.

packages/core/src/api/device/DeviceSettings.ts
wabicai: We are intentionally not restoring a mismatch error here. `DeviceSettingsPageShow -> Success` is the firmware-confirmed completion result for this device-side operation. The refresh never applies the requested value optimistically: it stores only fields returned by the device. When the device locks and omits the private passphrase field, the mapper/store preserves the last confirmed value; when an unlocked read returns the old value, that device-returned value remains authoritative in state. Returning `RuntimeError` or `DeviceCheckPassphraseStateError` would conflict with the App contract for this page flow. `4c0ded08a` adds an integration regression proving the locked-field omission preserves confirmed state.

wabicai:
Latest review follow-up is in `4c0ded08a` and versioned as `1.2.0-alpha.133` in `b6d3307b4`:

- A shared React Native BLE manager reset now removes subscriptions, rejects Protocol V2 readers, and emits an explicit disconnect for every cached session before clearing manager state. The multi-device regression verifies an unrelated session cannot remain silently stale.
- Wallpaper post-apply refresh is best-effort after `DeviceSettingsSet` succeeds, with a regression covering read-back failure.
- Passphrase page success remains non-optimistic and non-throwing by design; a real `Device`/StateStore regression verifies locked status preserves the last confirmed passphrase value.

Focused validation: 3 suites passed, 333 tests passed; changed-file ESLint and Prettier passed; package versions are aligned at `.133`.

packages/core/src/api/device/DeviceSettings.ts
```diff
@@ -201,20 +200,7 @@ export default class DeviceSettings extends BaseMethod<ApplySettings> {
           const res = await this.device.commands.typedCall('DeviceSettingsPageShow', 'Success', {
```
originalix, line 200: P1: [Passphrase settings can report success without changing the device]

This branch now ignores the post-page read-back result. If `DeviceSettingsPageShow` returns `Success` but the refreshed `status.passphraseProtection` is still the old value, the method still resolves success.

That means a user can exit the on-device page without actually enabling passphrase, and the SDK will still report that the setting succeeded. For a security-sensitive toggle, callers need a trustworthy success signal before they persist state or assume hidden-wallet flows are available.

Please keep the relaxed handling only for the locked-after-disable case, but still reject or surface a no-op when the refreshed state is authoritative and does not match the requested value.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1840,9 +1930,37 @@ export default class ReactNativeBleTransport {
   private resetPlxManager() {
```
originalix, line 1930: P2: [BLE manager reset leaves active Protocol V1 calls unresolved]

`runNativeTeardown()` can now reset the shared BLE manager after a single cleanup timeout. In `resetPlxManager()` we clear cached transports and destroy the manager, but we never reject `this.runPromise` / `this.runPromiseDeviceId`.

If another React Native BLE session is in the middle of a Protocol V1 call, that call is still waiting on its deferred after the underlying manager and subscriptions have already been torn down. The result is a long stall on an unrelated device, or an indefinite hang for flows that do not carry a timeout.

Please fail the active run promise as part of the manager reset, or funnel the reset through a disconnect path that settles in-flight Protocol V1 calls before the shared manager is destroyed.

packages/core/src/api/device/DeviceSettings.ts
wabicai: This is not a no-op success path for Pro2 firmware. `DeviceSettingsPageShow(DevicePassphrase)` returns `Failure` with `DeviceError_ActionCancelled` when the user cancels or abandons the page, and returns `Success` only through `foreground_device_control_page_show_confirmed()` after the user confirms the toggle. The state request also performs an immediate foreground-cache write-through before it is synchronized to the background state. When disabling passphrase locks the device, the locked `DeviceStatus` intentionally cannot expose `passphrase_enabled`; treating that unavailable read-back as a mismatch error would turn a confirmed successful disable into a false SDK failure. The SDK therefore refreshes device state without optimistic mutation and preserves the last confirmed passphrase value while the field is unavailable. We should not add a mismatch exception here.

packages/core/src/api/device/DeviceSettings.ts
```diff
@@ -201,20 +200,7 @@ export default class DeviceSettings extends BaseMethod<ApplySettings> {
...
-            updated.status.passphraseProtection !== requestedPassphrase &&
-            !lockedAfterDisabling
-          ) {
-            throw TypedError(
-              HardwareErrorCode.RuntimeError,
-              'Protocol V2 passphrase setting did not reach the requested value.'
-            );
-          }
+          await refreshStatusAndSettings();
           return res.message;
```
sidmorizon, line 204: **Turning the device passphrase option on or off reports success even when the device never changed it**

Severity: `non-severe`

The passphrase change is reported as successful (`return res.message` at `packages/core/src/api/device/DeviceSettings.ts:204`) without checking the value read back from the device, so a change the user declined on the device screen still looks like it worked.
Impact: An app can tell the user the passphrase option was switched when the device still has the old value.

<details>
<summary>Removed post-toggle verification, and inconsistency with the air-gap page flow</summary>

Previously this branch re-read `status`/`settings` and threw `Protocol V2 passphrase setting did not reach the requested value.` when `updated.status.passphraseProtection !== requestedPassphrase`, with an explicit allowance for the "device locked itself after disabling" case. The new code only calls `await refreshStatusAndSettings()` and discards the result (`packages/core/src/api/device/DeviceSettings.ts:203-204`).

The air-gap page branch immediately below still verifies the read-back and throws when the target was not reached (`packages/core/src/api/device/DeviceSettings.ts:215-221`), so the two on-device page flows now report failures differently for the same class of user action (user cancels on the device).

If the intent is that page-based flows only guarantee "page accepted" (per `docs/sdk/events.md`), the air-gap branch should be aligned the same way and the relaxed contract documented for `usePassphrase`; otherwise the passphrase verification (with the locked-device exception that the new tests cover) should be restored.

</details>

originalix (approved):

PR #883, opened 2026-08-17
fix: harden Pro2 device management, firmware, and cancellation

wabicai:
## Summary

- allow either the main PIN or an Attach PIN to authorize Pro2 device-management operations
- restore the Pro2 portfolio test route, generated fixtures, and regression cases
- stabilize Protocol V2 firmware-install polling and stop polling after cancellation
- send device-level Protocol V2 `Cancel` over WebUSB when cancellation arrives before a prompt callback is registered
- publish the changes as `1.2.0-alpha.135`

## Root causes

- Protocol V2 non-wallet methods defaulted to Main PIN pre-unlock unless they explicitly selected another PIN policy.
- Firmware polling could continue after cancellation or resolve from stale state.
- WebUSB cancellation could terminate the host request without sending `Cancel` to the device when no prompt callback had been registered yet.
- The Pro2 portfolio workflow lacked an executable UI route and complete fixture coverage.

## Changes

- use the Any PIN pre-unlock policy for PIN changes, device wipe, verification, and firmware updates
- add the Pro2 portfolio test page, fixture builder, manifest, signed packages, and focused tests
- tighten firmware V4 install polling lifecycle and cancellation cleanup
- extend Protocol V2 fallback cancellation from BLE to browser and desktop WebUSB
- update workspace package versions for the alpha release

## Impact

- Pro2 device-management operations can be authorized with either valid PIN type.
- Canceling a WebUSB signing or device operation returns the device to its normal flow instead of only rejecting the host promise.
- Firmware cancellation no longer leaves polling active.
- Protocol V1 behavior remains unchanged.

## Validation

- Protocol V2 unlock-policy tests passed
- device lifecycle cancellation tests: 34 passed
- focused firmware polling tests passed
- focused ESLint passed
- `@onekeyfe/hd-core` build passed
- GitHub build and security checks passed


packages/core/src/api/device/DeviceVerify.ts
```diff
@@ -11,7 +12,15 @@ import type { BixinVerifyDeviceRequest } from '@onekeyfe/hd-transport';
 import type { DeviceVerifySignature } from '../../types';
 
 export default class DeviceVerify extends BaseMethod<BixinVerifyDeviceRequest> {
+  getSupportedProtocols() {
+    return ['V1', 'V2'] as const;
```
originalix, line 16: P1: [Protocol V2 verification still uses V1-only message names]

When `deviceVerify` is invoked on a Protocol V2 device, this declaration routes `run()` into `SESignMessage` and `ReadSEPublicCert`. Those legacy types have no `MessageType_*` entries in `messages-protocol-v2.json`, so the V2 serializer throws before the verification request reaches the device.

Branch on Protocol V2 and use `DeviceCertificateSign -> DeviceCertificateSignature` and `DeviceCertificateRead -> DeviceCertificate`, map their fields to the public `{ cert, signature }` result, and add a test that executes `run()` through the V2 schema.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1623,9 +1629,11 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
               Math.floor(statusVersion / 0x100) % 0x100
             }.${statusVersion % 0x100}`;
       if (!observedVersion) {
-        if (this.protocolV2FinalStatusVerified) {
+        const isUnobservableApplicationSlot =
+          applicationTargets.length > 1 && (target === 'app_v1' || target === 'app_v2');
```
originalix: P1: [Multi-slot fallback can report an unverified firmware install as complete]

When both app slots are requested and `DeviceFirmwareUpdateStatusGet` disappears after reboot, the fallback can return after any matching install evidence or any observable app-version change. At this point `protocolV2FinalStatusVerified` is still false and neither per-slot version is observable, but this condition suppresses both missing-version errors solely because two app targets were requested. A partial install—one slot changed while the other never completed—can therefore be reported as successful.

Only accept unobservable app-slot versions when all requested target IDs were proven complete, or preserve equivalent per-slot completion evidence across the reboot; target count alone must not bypass final verification.

originalix:

@wabicai

Reviewed commit `fae9cf5fd2f2`. This report always reflects the latest reviewed changes.

### Review summary

Core remains responsible for Protocol V2 device-management authorization, unified device state, certificate verification, cancellation, release selection, and firmware orchestration. The PR allows either the main or Attach PIN for selected Pro2/Neo management calls, adds Protocol V2 certificate verification, preserves private settings across restricted reads, expands WebUSB cancellation, and revises multi-target firmware installation and verification. The Expo playground also gains a generated Portfolio fixture harness, while package versions are aligned at alpha.142.

**What needs attention:** Hold the merge until firmware completion is proven separately for every requested target and locked-state transitions clear the active Attach-PIN flag. Then validate the broadened management authorization and firmware paths on physical Pro2/Neo devices over USB and BLE.

### Issues to address

- **P1 · Do not promote partial firmware evidence to all targets**
  - **Impact:** When a multi-target installation returns to App mode, a generic status acknowledgement or a version change in any one target can satisfy the fallback completion branch. The helper then marks every requested target complete, and final verification accepts missing target versions based on that set. A component that was skipped or failed without a retained status record can therefore be reported as successfully installed.
  - **Suggested action:** Track completion evidence separately for every target and add an ID only after that target reports a finished status or its own expected version is observed. Reject every requested target lacking individual evidence.
- **P2 · Locked status can retain a stale Attach-PIN unlock flag**
  - **Impact:** When a device previously unlocked through an Attach PIN later reports `unlocked: false`, the mapper omits `unlockedAttachPin`. `DeviceStateStore` ignores undefined patch values, so the previous `true` survives and is projected back into compatibility features. Public state becomes contradictory, and wallet-session checks can misclassify the locked context as Attach-PIN unlocked.
  - **Suggested action:** Map `unlockedAttachPin` to `false` whenever `status.unlocked === false`, preserve only configuration fields that are genuinely unavailable while locked, and add a store-level unlocked-to-locked transition test.

### Validation gaps

- `yarn check-versions` passed for all 30 packages at `1.2.0-alpha.142`.
- The Portfolio generator passed Python syntax validation; its manifest has 40 unique cases, exactly matches the 39 committed packages, and all packages remain below 27 KB.
- Focused Jest, TypeScript, and ESLint checks could not run because their workspace executables are not installed.
- The committed signed Portfolio packages were not regenerated from the sibling firmware and application repositories.
- PIN changes, wipe, certificate verification, cancellation, and multi-target firmware installation were not exercised on physical Pro2/Neo devices over USB and BLE.

packages/core/src/api/device/DeviceVerify.ts
originalix, line 16: P1: [Protocol V2 verification still uses V1-only message names]

When `deviceVerify` runs on a Protocol V2 device, this declaration routes `run()` into `SESignMessage` and `ReadSEPublicCert`. Those legacy names have no `MessageType_*` entries in the V2 schema, so `createMessageFromName()` throws before the verification request reaches the device.

Add a V2 branch using `DeviceCertificateSign -> DeviceCertificateSignature` and `DeviceCertificateRead -> DeviceCertificate`, map `data` and `cert_and_pubkey` into the existing public result, and cover the real V2 serialization path in a test.

packages/core/src/api/FirmwareUpdateV4.ts
originalix: P1: [Multi-slot fallback can report an unverified firmware install as complete]

When both app slots are requested and `DeviceFirmwareUpdateStatusGet` disappears after reboot, polling can return after any matching install evidence or any observable app-version change. At this point `protocolV2FinalStatusVerified` can still be false and neither per-slot version is observable, but this condition suppresses both missing-version errors solely because two app targets were requested. A partial install can therefore be reported as successful.

Only accept unobservable app-slot versions after every requested target ID has explicit completion evidence, or preserve equivalent per-slot evidence across the reboot.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2499,14 +2525,9 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
       Log.error(`[FirmwareUpdateV4] firmware install failed target=${failedTargetDetails}`);
-      throw ERRORS.TypedError(
-        HardwareErrorCode.FirmwareError,
-        `Protocol V2 firmware target failed: target=${failedTarget.target_id} status=${
-          failedTarget.status ?? 'unknown'
-        } payloadVersion=${failedTarget.payload_version ?? 'unknown'} path=${
-          failedTarget.path ?? 'unknown'
-        }`
-      );
+      throw ERRORS.TypedError(HardwareErrorCode.FirmwareError, undefined, {
```
originalix: P2: [Protocol V2 install failures lose actionable public diagnostics]

This branch used to emit a distinct error message for the failed install record, but it now throws a generic `FirmwareError` and keeps the real reason only in `firmwareUpdateCode`.

The problem is that the public error path only forwards `error.message` and `errorCode`; it does not expose `error.params`. In practice that means timeout, target-conflict, and device-reported install failure now all surface as the same user-visible failure string, so callers and support logs lose the information needed to tell which install stage actually broke.

Please keep a branch-specific message here, or plumb `firmwareUpdateCode` through the public error payload before collapsing these failures into the generic `FirmwareError` text.

packages/connect-examples/expo-playground/public/portfolio-cases/manifest.json
```diff
@@ -0,0 +1,2848 @@
+{
+  "version": 2,
+  "firmwareCommit": "9d58ee244f77324f45683e8d78f821c00ae095dc",
+  "appCommit": "0c4008b7491862d6a4d9c4809cbeea8c0db377d7",
+  "sdkCommit": "4501ad9a9c46eb74237e0f69e4c8bc1d1b02b365",
```
originalix, line 5: P3: [Committed portfolio manifest provenance is stale]

This manifest says the fixtures were generated from SDK commit `4501ad9a9c46eb74237e0f69e4c8bc1d1b02b365`, but the checked-out PR head is `15e3f0dc6c543c83a96e63b5d19aad70cdd6efb1`.

The page uses this metadata to explain which firmware/app/SDK revisions produced the committed packages, so leaving `sdkCommit` behind means the new harness will point reviewers and testers at the wrong SDK revision when they try to debug a fixture mismatch.

Please regenerate the manifest from the current head, or stop serializing `sdkCommit` if these assets are intentionally kept stable across later code changes.

packages/connect-examples/expo-playground/app/routes/portfolio-test.tsx
```diff
@@ -0,0 +1,534 @@
...
+    <PageLayout>
+      <div className="space-y-5 p-4 sm:p-6" data-testid="portfolio-test-page">
+        <div className="flex flex-wrap items-start justify-between gap-4">
+          <div className="max-w-3xl">
+            <div className="flex items-center gap-2">
+              <ShieldCheck className="h-7 w-7 text-primary" />
+              <h1 className="text-2xl font-semibold text-foreground">Pro2 Portfolio 专项测试</h1>
+            </div>
+            <p className="mt-2 text-sm leading-6 text-muted-foreground">
+              逐项验证金额格式、Native/Contract/All Networks Token 映射、精确 JSON 结构与固件边界。
```
originalix: P3: [Portfolio test page copy is inconsistent with the manifest and existing i18n setup]

This description says batch mode runs every 15 seconds, but the committed manifest sets `intervalMs` to `60000`, and the badge a few lines below already renders a 60-second interval.

At the same time, the page body is hard-coded in Chinese even though this change already added locale entries for the navigation label. In English builds that leaves the new page half localized, and the operator guidance disagrees with the actual runner timing.

Please derive the interval wording from `manifest.intervalMs` and move the page copy through the existing locale files so the UI stays consistent with runtime behavior.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Addressed in `e038b075f`. The fallback now accepts unobservable versions only when every requested install target ID is present in `protocolV2CompletedTargetIds`. A regression test verifies that completion evidence for only one app slot is rejected.

packages/core/src/api/device/DeviceVerify.ts
wabicai: Addressed in `e038b075f`. `DeviceVerify` now branches on Protocol V2, uses `DeviceCertificateSign` / `DeviceCertificateRead`, maps the V2 response fields to the public result, and is covered by the real V2 typed-call path in `protocol-v2-unlock-policy.test.ts`.

wabicai: Addressed in `e038b075f`. `DeviceVerify` now uses the Protocol V2 certificate message names and maps `data` / `cert_and_pubkey`; the V2 serialization path is covered by `protocol-v2-unlock-policy.test.ts`.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Addressed in `e038b075f`. Multi-slot fallback requires explicit completion evidence for all requested target IDs; partial per-slot evidence no longer bypasses final verification.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Addressed in `c2411d2fe`. Failed target status, status conflict, and timeout now have distinct public messages while preserving `params.firmwareUpdateCode`. The focused Protocol V2 suite passes (281 tests).

packages/connect-examples/expo-playground/public/portfolio-cases/manifest.json
wabicai: No code change is needed here. `sdkCommit` records the SDK checkout actually used by `build_fixtures.py` when the signed fixtures are generated; it is provenance for the generated inputs, not the eventual PR head. The generator reads `git rev-parse HEAD` before the generated manifest can be committed, so making it equal to the commit that contains the manifest would be self-referential. `4501ad9...` is therefore the correct generation input for these committed packages.

packages/connect-examples/expo-playground/app/routes/portfolio-test.tsx
wabicai: Addressed in `c2411d2fe`. Static page copy and result labels now use the existing English/Chinese i18n setup, and both the description and badge derive the interval from `manifest.intervalMs`. Playground typecheck, ESLint, and the 11 Portfolio tests pass.

packages/core/src/device/DeviceStateMapper.ts
```diff
@@ -347,8 +347,10 @@ export const mapProtocolV2DeviceStatusToState = (status: DeviceStatus): DeviceSt
     unlocked: status.unlocked,
     passphraseProtection: status.unlocked === true ? status.passphrase_enabled ?? null : undefined,
     backupRequired: status.backup_required,
-    attachToPinEnabled: status.attach_to_pin_enabled,
-    unlockedAttachPin: status.unlocked_by_attach_to_pin,
+    attachToPinEnabled:
+      status.unlocked === true ? status.attach_to_pin_enabled ?? undefined : undefined,
+    unlockedAttachPin:
```
originalix, line 352: P2: [Locked status can retain a stale Attach-PIN unlock flag]

When the device was previously unlocked through an Attach PIN and a later status reports `unlocked: false`, this branch emits `undefined` for `unlockedAttachPin`. `DeviceStateStore` treats `undefined` as “leave the old value unchanged,” so the prior `true` survives and Core can publish `unlocked: false` together with `unlockedAttachPin: true`.

That stale runtime flag can mislead wallet-session routing and consumers of DeviceState. Please explicitly clear `unlockedAttachPin` when the device becomes locked, while continuing to omit only settings that are genuinely unreadable in the locked state, and add a store-level transition test for Attach-PIN-unlocked to locked.

packages/core/src/device/DeviceStateMapper.ts
```diff
@@ -347,8 +347,10 @@ export const mapProtocolV2DeviceStatusToState = (status: DeviceStatus): DeviceSt
...
+      status.unlocked === true ? status.unlocked_by_attach_to_pin ?? null : undefined,
```
originalix, line 353: P2: [Locked status can retain a stale Attach-PIN unlock flag]

When the device was previously unlocked through an Attach PIN and a later status reports `unlocked: false`, this branch emits `undefined` for `unlockedAttachPin`. `DeviceStateStore` treats `undefined` as “leave the old value unchanged,” so the previous `true` survives and Core can publish `unlocked: false` together with `unlockedAttachPin: true`.

Please explicitly set this runtime flag to `false` when the device is locked, while continuing to preserve only configuration fields that are genuinely unavailable in the locked state. A store-level transition test from Attach-PIN-unlocked to locked would cover the regression.

originalix, line 353: P2: [Locked status can retain a stale Attach-PIN unlock flag]

When the device was previously unlocked through an Attach PIN and a later status reports `unlocked: false`, this branch emits `undefined` for `unlockedAttachPin`. `DeviceStateStore` treats `undefined` as “leave the old value unchanged,” so the previous `true` survives and Core can publish `unlocked: false` together with `unlockedAttachPin: true`.

Please explicitly set this runtime flag to `false` when the device is locked, while continuing to preserve only configuration fields that are genuinely unavailable in the locked state. Add a store-level transition test from Attach-PIN-unlocked to locked.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2822,26 +2841,15 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
               Log.log(
                 '[FirmwareUpdateV4] firmware status endpoint unavailable after confirmed App reboot'
               );
+              this.recordProtocolV2AuthoritativeInstallCompletion(expectedTargetIds);
```
originalix, line 2812: P1: [Do not promote partial firmware evidence to all targets]

When a multi-target install returns to App mode after only a generic acknowledgement or a version change in one target, this branch marks every expected ID complete. Final verification then treats those IDs as authoritative when a target version is missing, so a component that never finished can be reported as installed.

Please preserve completion evidence per target. Only add an ID after that target reports a finished status or its own expected version is observed, and reject any requested target without individual evidence.

packages/core/src/device/DeviceStateMapper.ts
originalix, line 353: P2: [Locked status can retain a stale Attach-PIN unlock flag]

When a device previously unlocked through an Attach PIN later reports `unlocked: false`, this conditional omits `unlockedAttachPin`. `DeviceStateStore` ignores undefined patch values, so the previous `true` survives and is projected back into compatibility state, leaving a locked device classified as Attach-PIN unlocked.

Please map `unlockedAttachPin` to `false` whenever the device is locked, while preserving only configuration fields that are genuinely unavailable, and add an unlocked-to-locked store transition test.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1965,6 +1987,10 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         ? PROTOCOL_V2_UPDATE_TARGET_BY_TARGET_ID.get(target.targetId)
         : undefined;
       if (updateTarget && targetsToUpdate.has(updateTarget)) {
+        if (component.version) {
+          this.params.expectedTargetVersions ??= {};
+          this.params.expectedTargetVersions[updateTarget] = component.version.join('.');
+        }
```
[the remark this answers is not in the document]
wabicai: Resolved by `2d57eb2f0`. `prepareRemoteProtocolV2Binaries` now assigns the release-config expected version only when the target is actually downloaded, so an explicit artifact preserves its caller-provided expectation. The regression test `downloads only firmware targets that are missing from an explicit mixed payload` keeps `app_v1: 9.9.9` while assigning the downloaded coprocessor version.

packages/connect-examples/expo-playground/app/routes/portfolio-test.tsx
```diff
@@ -0,0 +1,534 @@
...
+              批量模式每隔 15 秒执行一次；正向用例必须完整通过上传和 PortfolioUpdate，反向用例必须返回指定拒绝原因。
```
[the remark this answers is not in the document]
wabicai: Resolved by `c2411d2fe`. The page now derives `intervalSeconds` from `manifest.intervalMs` and uses that value in both the description and interval badge, so the displayed interval stays aligned with the committed manifest.

packages/core/src/device/DeviceStateMapper.ts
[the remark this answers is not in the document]
wabicai: Duplicate of the latest unresolved Attach-PIN transition thread (PRRT_kwDOHV2MTM6Z2RIQ). Keeping that thread open as the canonical actionable item; no separate code change is needed for this duplicate.

packages/core/src/device/DeviceStateMapper.ts
wabicai: Duplicate of the latest unresolved Attach-PIN transition thread (PRRT_kwDOHV2MTM6Z2RIQ). Keeping that thread open as the canonical actionable item; no separate code change is needed for this duplicate.

packages/core/src/device/DeviceStateMapper.ts
wabicai: Duplicate of the latest unresolved Attach-PIN transition thread (PRRT_kwDOHV2MTM6Z2RIQ). Keeping that thread open as the canonical actionable item; no separate code change is needed for this duplicate.

wabicai: Duplicate of the latest unresolved Attach-PIN transition thread (PRRT_kwDOHV2MTM6Z2RIQ). Keeping that thread open as the canonical actionable item; no separate code change is needed for this duplicate.

wabicai:
Product decisions for the remaining review threads:

1. **Protocol V2 multi-target completion** — Installation is treated as reliable and atomic. Once the device acknowledges the installation and returns to App mode, that transition is authoritative completion for the requested target set. Per-target observable versions are not required for this fallback, so promoting the requested targets is intentional.

2. **Locked Attach-PIN compatibility state** — `status.unlocked` is the authoritative lock state. While locked, Attach-PIN detail may be unavailable, and preserving the prior compatibility field is accepted by the current product behavior. Consumers must gate locked/unlocked behavior on `status.unlocked`; no change is required in this scope.

3. **Portfolio Pro2 selection** — The route is an Expo playground/example-only physical-device test harness, not part of the production App or SDK workflow. Multi-device selection is outside the accepted scope; operators run it with one intended Pro2 connected.

No SDK code changes are required for these three threads.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -90,9 +91,8 @@ const PROTOCOL_V2_BOOTLOADER_RECONNECT_TIMEOUT = 90 * 1000;
 const PROTOCOL_V2_FINAL_RECONNECT_TIMEOUT = 3 * 60 * 1000;
 const PROTOCOL_V2_SHORT_RESPONSE_TIMEOUT = 5 * 1000;
 const PROTOCOL_V2_FIRMWARE_STATUS_RESPONSE_TIMEOUT = 15 * 1000;
```
sidmorizon, line 93: **Firmware updates can be aborted while the device is still installing**

Severity: `non-severe`

The overall waiting window for a Pro 2 firmware installation was shortened from eight minutes to five (`PROTOCOL_V2_INSTALL_TIMEOUT` at `packages/core/src/api/FirmwareUpdateV4.ts:93`), and that same window also has to cover the time the user spends confirming the update on the device, so a slow confirmation plus a long multi-part install now ends in a reported failure while the device keeps installing.
Impact: Users updating several firmware components, or who take a while to confirm on the device, can see the update fail even though the device is still writing firmware.

<details><summary>Why the shortened window now covers confirmation and every install target</summary>

`executeProtocolV2TransferPhase` (`packages/core/src/api/FirmwareUpdateV4.ts:2379-2396`) shows the "confirm on device" tip, sends the install request and then immediately enters `waitForProtocolV2FirmwareUpdateComplete`. The poll loop condition is `Date.now() - startTime < PROTOCOL_V2_INSTALL_TIMEOUT` (`packages/core/src/api/FirmwareUpdateV4.ts:2676`) and one of its wait states is explicitly "waiting for user confirmation or target status" (`packages/core/src/api/FirmwareUpdateV4.ts:2772-2774`), so on-device confirmation time, bootloader reboots, reconnect retries and the sequential install of every requested target (bootloader, app P1/P2, coprocessor, SE01–SE04) all share the same five-minute budget. When it elapses the method throws `FirmwareError` with `firmwareUpdateCode: 'FirmwareInstallTimeout'` (`packages/core/src/api/FirmwareUpdateV4.ts:2861-2867`) even though the device is unaffected and continues installing.

The change also removed the separate 30s grace windows, so the total budget is now the only safety net; keeping the previous 8-minute (or a larger) total window would preserve the old tolerance while retaining the new, simpler polling logic.

</details>

Suggested fix:

```ts
const PROTOCOL_V2_INSTALL_TIMEOUT = 8 * 60 * 1000;
```

originalix (approved):

PR #884, opened 2026-08-17
fix: cache WebUSB protocol probe and merge V1 identity check into Initialize (OK-59992)

originalix:
## Problem

OK-59992: after upgrading the SDK from 1.1.31 to 1.2.0-alpha.x, every hardware call on Protocol V1 devices (Pro/Classic/Mini) over WebUSB became ~2x slower (measured +1.7–1.9s per call), making multi-call flows such as hidden-wallet creation 50%+ slower overall.

## Root causes (one commit each)

1. **Wire-level protocol probe on every acquire**: `hd-transport-web-device`'s `acquire()` sent a bare `Initialize` probe on every call (`expectedProtocol='V1'` only set the probe order, it never skipped the probe), and `release()` dropped the probe cache — while core releases the session after every API call. Every SDK call paid a full probe round trip plus a USB close/reopen cycle.
2. **GetFeatures preflight in `Device.initialize`**: every deviceId-carrying call (all derivation/signing methods) issued a read-only `GetFeatures` identity check before the actual `Initialize` — two round trips doing one round trip's work (+~0.9s per call).

## Changes

**`fix: cache WebUSB protocol probe across acquires`**
- Probe results are cached per USB serial and survive `release()`; first contact still probes as before
- USB `disconnect` events and transfer-level reconnects only MARK the cache stale (the next acquire re-probes from scratch). This covers the reboot/mode-change case — the only way a device's protocol can change, and it always surfaces as a disconnect — while preserving the pre-existing transfer-retry recovery for transient re-enumerations mid-session. A missed disconnect event self-heals through the transfer-error path
- Protocol V2 behavior is unchanged: V2 entries are still dropped through the existing link-invalidation callback (the link-manager early-return path was verified)
- New `acquiredPaths` session gate: with the protocol cache surviving release, `deviceProtocol` can no longer double as a session marker, so post-release straggler `call()`/`post()` now fail fast instead of silently reopening the device
- The disconnect listener is module-scoped and attached once (a per-instance listener would leak every transport instance across SDK re-initializations)

**`fix: validate device identity from the Initialize round trip`**
- Removes the GetFeatures preflight; `checkDeviceId` now runs on the single `Initialize` response (which is itself a Features message carrying `device_id`). The method fn has not run yet, so a mismatch still fails before any wallet data can be derived, with the same `DeviceCheckDeviceIdError`
- Preserves one load-bearing side effect of the old preflight: `TransportManager.reconfigure(this.features)` runs before encoding Initialize (local operation, no wire I/O), so a stale legacy schema (old-firmware Touch/Mini on multi-device setups) cannot silently strip `passphrase_state` / `is_contains_attach` from the wire message
- Both the mismatch path and the reconfigure-failure path clean up any session the wrong device's Initialize may have cached

## Measurements (desktop, same OneKey Pro, same UX flow, silent-call totals)

| Phase | 6.5.0 baseline (1.1.31) | Regression (alpha.129) | This PR (alpha.134 verification build) |
|---|---|---|---|
| Onboarding wallet creation | 14.0s | 25.6s | **14.7s** |
| Enable passphrase + create hidden wallet | 18.1s | 27.2s | **18.9s** |

Per call: btcGetPublicKey 3.75s → **1.88s**, evmGetAddress 3.65s → **1.87s**, getFeatures 1.2s → **0.21s** — all back to the 1.1.31 baseline. Pro2 (V2): no regression on any metric.

## Verification

- 12 new tests (probe-cache lifecycle, stale semantics, session gate, single-round-trip identity check, error-path cleanup, schema resync ordering); all affected core + transport suites green
- Two adversarial multi-agent review rounds (27 agents); both real defects found (legacy-schema silent field stripping, missing error-path cleanup) are fixed with test coverage
- This PR carries the version bump to `1.2.0-alpha.137` (probed free on npm at commit time). **Release flow: merge into `onekey` first, then dispatch `package-publish.yml` from `onekey`** — the workflow reads the version from `packages/core/package.json`. If a parallel pipeline takes alpha.137 before the merge lands, re-probe and bump on `onekey` before dispatching (`lerna publish from-package` silently skips already-published versions, so verify the published tarball content afterwards)
- Verification builds used throwaway versions `1.2.0-alpha.134`/`.136` (dist-tag `ok59992`); they are not part of this PR


packages/hd-transport-web-device/src/webusb.ts
```diff
@@ -231,20 +276,36 @@ export default class WebUsbTransport extends ProtocolV2UsbTransportBase<string>
...
+        this.deviceProtocol.delete(input.path);
       }
-      await this.detectProtocol(input.path, input.expectedProtocol, protocolHint);
+      const cachedProtocol = this.deviceProtocol.get(input.path);
+      if (cachedProtocol && input.expectedProtocol && cachedProtocol !== input.expectedProtocol) {
+        // The caller expects a different protocol than the cached probe result;
+        // the cache is stale — drop it and re-probe on the wire below.
+        this.deviceProtocol.delete(input.path);
+      }
+      if (!this.deviceProtocol.has(input.path)) {
```
originalix, line 318: P1: [Cached V1 results bypass active protocol detection]

After release retains a V1 entry, the next acquire opens a new USB connection but this condition skips `detectProtocol()` whenever the cached value matches the expectation. If the device rebooted or changed mode and disconnect invalidation was delayed, missed, or routed elsewhere, the new link is still treated as V1 and receives V1 framing even when it is now V2.

Keep the cached value only as the strict expectation or probe-order hint, and always validate that expectation with an active response after connecting.

packages/core/src/device/Device.ts
```diff
@@ -960,7 +949,46 @@ export class Device extends EventEmitter {
...
 
-      await callInitialize(payload, options?.initSession);
+      if (this.features) {
+        // Re-sync the V1 message schema for THIS device before encoding
+        // Initialize: the process-global schema may still reflect another
+        // device (e.g. legacy-firmware Touch/Mini) on multi-device setups, and
+        // a stale legacy schema would silently strip passphrase_state /
+        // is_contains_attach from the wire message. Local operation, no wire
+        // I/O; a no-op when the schema is unchanged.
+        await TransportManager.reconfigure(this.features);
```
originalix, line 977: P2: [Schema resynchronization is not atomic with Initialize]

When two Protocol V1 devices requiring different schemas initialize concurrently, this call mutates process-global transport schema state but does not hold that state through the following `typedCall`. The other device can switch the schema in between, so this Initialize may be encoded with the wrong schema and silently omit fields such as `session_id` or `passphrase_state`.

Serialize the schema switch and its matching Initialize encoding in one transport-scoped critical section, and add a concurrent two-device regression test.

originalix:

@originalix

Reviewed commit `9e97c8e8b76c`. This report always reflects the latest reviewed changes.

### Review summary

The PR adds a persistent WebUSB protocol-probe cache across release/reacquire cycles, introduces stale-cache invalidation for disconnect and retry paths, refactors V1 initialization to validate identity from Initialize responses instead of an unconditional GetFeatures preflight, expands unit coverage around those flows, and bumps the monorepo package versions to 1.2.0-alpha.137.

**What needs attention:** Re-check the V1 identity-proof path across any release, reconnect, or transport-disconnect boundary; the WebUSB cache logic is focused and well tested, but the core session-safety invariant regressed there.

### Issues to address

- **P1 · V1 initialize now trusts stale features as proof of device identity**
  - **Impact:** Before this change, every `deviceId`-guarded V1 initialize performed a live `GetFeatures` read before any wallet context was sent. The new branch skips that live proof whenever `this.features` still matches `expectedDeviceId`, but V1 feature state survives both `release()` and `markTransportDisconnected()`. After a reconnect, wipe/recover, or same-path swap, the next `Initialize` can therefore send `session_id` and `passphrase_state` under a stale schema to an unverified device before the mismatch is detected.
  - **Suggested action:** Only skip the context-free identity preflight when the identity was observed on the current transport link, or invalidate V1 features across release/disconnect so cached state is never treated as live proof.

### Validation gaps

- Static review only for execution checks: `yarn test ...` falls through to `lerna run test ...`, but `lerna` is not installed in this workspace, so I could not run the targeted Jest suites.
- I did run `node scripts/check-versions.js`, which passed and confirms the package version roll is internally consistent.

packages/hd-transport-web-device/src/webusb.ts
originalix: Addressed in 20966e25e. The cache-hit path now validates the **USBDevice object identity** against the exact object the protocol was probed on: the browser keeps object identity stable while a device stays connected, and any replug/reboot yields a new object — so a delayed, missed, or mis-routed disconnect event can no longer let a stale entry through; the acquire re-probes on the wire whenever the identity changed. Combined with the existing guards (disconnect marking, transfer-retry stale marking, expectedProtocol conflict), every path where the physical protocol could have changed now forces an active probe, while the steady-state case (same object, same plug session — where the protocol physically cannot change) keeps the single-probe-per-plug behavior that this PR exists to deliver. Always re-probing after connect would reinstate the ~1-1.9s per-call regression from OK-59992.

packages/core/src/device/Device.ts
originalix: Agreed the race is real, but it is pre-existing and broader than this hunk: the removed preflight had the identical non-atomic shape (reconfigure → typedCall with no lock), and every other V1 call site encodes against the process-global schema without holding it (getFeatures at Device.ts, the initialize call itself in both versions, all method fns). This PR neither introduces nor widens the window — it moves the same resync one statement earlier so the Initialize encode at least starts from this device's schema, which the alpha.129 code only guaranteed for deviceId-carrying calls. A correct fix is schema-per-call (thread the parsed schema through TransportCallOptions instead of mutating the singleton), which touches every transport and is out of scope for this hotfix; serializing just this one reconfigure+Initialize pair would leave the same race open for every other message type while adding a cross-device lock on the hottest path. Happy to file a follow-up issue for the schema-per-call refactor if you agree with that direction.

packages/core/src/device/Device.ts
```diff
@@ -960,7 +949,46 @@ export class Device extends EventEmitter {
...
+          // under the same passphraseState — a deliberate conservative purge
+          // after a physical-swap event, consistent with
+          // reconcileDeviceIdentity purging the previous device's sessions.)
+          this.clearInternalState();
+          throw ERRORS.TypedError(HardwareErrorCode.DeviceCheckDeviceIdError);
+        }
+      };
+
+      try {
+        await callInitialize(payload, options?.initSession);
```
originalix, line 1001: P1: [Identity is checked only after wallet context is sent]

When `options.deviceId` refers to a cached wallet but another Protocol V1 device or wallet lifecycle is present at this path, this call sends the cached `session_id` and `passphrase_state` before checking the returned `device_id`. The later purge cannot undo disclosure of that internal session handle or the wallet-context operation already attempted on the wrong device.

Perform an identity-only Initialize without session or passphrase fields, verify its `device_id`, and only then load and send the cached wallet context.

packages/hd-transport-web-device/src/webusb.ts
```diff
@@ -231,20 +286,49 @@ export default class WebUsbTransport extends ProtocolV2UsbTransportBase<string>
...
+          this.deviceProtocol.delete(input.path);
+        }
+      }
+      const cachedProtocol = this.deviceProtocol.get(input.path);
+      if (cachedProtocol && input.expectedProtocol && cachedProtocol !== input.expectedProtocol) {
+        // The caller expects a different protocol than the cached probe result;
+        // the cache is stale — drop it and re-probe on the wire below.
+        this.deviceProtocol.delete(input.path);
+      }
+      if (!this.deviceProtocol.has(input.path)) {
```
originalix, line 318: P1: [Cached V1 results bypass active protocol detection]

After release retains a V1 entry, the next acquire opens a USB connection but this condition skips `detectProtocol()` when invalidation has not fired and the `USBDevice` object still matches. If a reboot or loader transition is missed or routed to another transport instance, the new connection is treated as V1 without an active response and can receive the wrong framing.

Keep the cached value only as the expected protocol or probe-order hint, and always validate it with `detectProtocol()` after connecting.

packages/core/src/device/Device.ts
originalix, line 977: P2: [Schema resynchronization is not atomic with Initialize]

When two Protocol V1 devices requiring different schemas initialize concurrently, this call mutates process-global transport schema state without holding it through the following `typedCall`. The other device can switch the schema in between, causing this Initialize to be encoded with the wrong schema and silently omit fields such as `session_id` or `passphrase_state`.

Serialize the schema switch and its matching Initialize encoding in one transport-scoped critical section, and cover the mixed-schema case with a concurrent regression test.

packages/core/src/device/Device.ts
originalix: The alternative proposed here — an identity-only Initialize, then a second wallet-context Initialize — is two wire round trips per deviceId-carrying call, which reinstates the exact per-call regression this PR exists to remove (OK-59992 measured it at ~0.9s per call on a Pro; the hidden-wallet flow pays it 3+ times). On the disclosure itself, the exposure is narrow and non-actionable: (1) `session_id` is an opaque handle scoped to the device that issued it — a different physical device cannot use, resume, or learn anything from it, it simply allocates a fresh session; (2) `passphrase_state` is a wallet fingerprint (a derived-address identifier), not key material — nothing can be derived from it; (3) the receiver in this scenario is another OneKey device the user physically plugged into their own machine, and Initialize itself derives nothing and signs nothing — the mismatch throws before any method fn runs, so no wallet DATA ever comes from or goes to the wrong device (same `DeviceCheckDeviceIdError` as the preflight version, plus the session purge and the later `checkPassphraseStateSafety` re-validation as additional layers). For reference, 1.1.31 shipped for years sending this exact payload with no identity check at all; this PR keeps #845's meaningful guarantee (no wallet data across devices) at single-round-trip cost. If the team judges the fingerprint disclosure itself unacceptable, that is a deliberate product tradeoff decision — happy to gate the preflight behind a config flag in a follow-up rather than re-paying the round trip unconditionally.

packages/hd-transport-web-device/src/webusb.ts
originalix: The object-identity check is deliberately event-independent, which is what closes the missed/mis-routed-event class: it never relies on the disconnect listener firing or being routed to the right instance. WebUSB binds a `USBDevice` object to a single connection instance — a reboot or loader transition tears down the USB connection, and the browser re-enumerates the device as a NEW object; there is no mechanism by which the same object identity survives a reboot, and there is no mechanism by which a device changes protocol WITHOUT rebooting (V1/V2 framing is firmware-mode-bound). So 'invalidation has not fired AND the USBDevice object still matches AND the device rebooted' is not a reachable state: if the device rebooted, the object in the refreshed `deviceList` differs and this condition re-probes on the wire. A mis-routed stale mark on a non-active transport instance is likewise covered, because each instance performs the identity comparison against its own refreshed device list, not against events. Always re-probing after connect would reinstate the ~1–1.9s per-call regression from OK-59992 for the steady-state case where the protocol physically cannot have changed.

packages/core/src/device/Device.ts
originalix: Filed #885 to track this properly with the schema-per-call design (thread the parsed schema through TransportCallOptions, resolved per device from its Features, keeping the global only as the pre-Features default) plus the concurrent mixed-schema regression test you describe. Keeping it out of this PR deliberately: the race predates this change (the alpha.129 preflight had the identical non-atomic reconfigure → typedCall shape, and every other V1 call site encodes against the global schema without holding it), this PR only narrows the practical window by moving the resync adjacent to the encode, and serializing just this one reconfigure+Initialize pair would leave the same race open for every other message type while adding a cross-device lock on the hottest path of a perf hotfix.

wabicai (dismissed):

packages/core/src/device/Device.ts
originalix, line 1001: P1: [Identity is checked only after wallet context is sent]

When `options.deviceId` identifies a cached wallet but a different Protocol V1 device is present at this path, this call sends the cached `session_id` and `passphrase_state` before checking the returned `device_id`. Purging the cache afterward cannot undo disclosure of the internal session handle or the wallet-context operation already attempted on the wrong device.

Perform an identity-only `GetFeatures` exchange first, verify its `device_id`, and only then load and send the cached wallet context.

packages/core/src/device/Device.ts
originalix, line 977: P2: [Schema selection is not atomic with Initialize encoding]

When two Protocol V1 devices requiring different schemas initialize concurrently, this call mutates process-global transport schema state without holding it through the following `typedCall`. The other device can switch the schema in between, causing this Initialize to omit fields such as `session_id` or `passphrase_state`.

Serialize the schema switch and its matching Initialize encoding in one transport-scoped critical section, and cover the mixed-schema case with a concurrent regression test.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/core/src/device/Device.ts
originalix: Fixed in f518b132f with an evidence-gated approach that keeps the hot path at one round trip: when `options.deviceId` is set but there is NO locally-cached evidence that the device at this path is the expected one (`this.features` missing, or its device_id already disagreeing with the caller — i.e. first contact and exactly the ambiguous cases where the disclosure risk lives), initialize now sends a **context-free Initialize first** (`{is_contains_attach: true}` only — no session_id, no passphrase_state), verifies `device_id` from its response, and only then issues the wallet-context Initialize. When cached features already match the expected id (the steady state in real flows — enumerate/getFeatures always run first and every call response refreshes features), the single-round-trip context Initialize is kept, with the post-response check as the backstop for the residual stale-features window. Regression tests added: an unproven device receives exactly one context-free Initialize and never sees the cached session_id/passphrase_state; the happy first-contact path proves identity then resumes the cached session.

packages/core/src/device/Device.ts
originalix: Tracked in #885 (filed from the previous round of this finding): per-call schema selection threaded through TransportCallOptions plus the concurrent mixed-schema regression test. The race predates this PR (the alpha.129 preflight had the identical non-atomic reconfigure → typedCall shape, and every V1 call site encodes against the process-global schema without holding it); this PR only narrows the window by moving the resync adjacent to the encode. Keeping the cross-cutting fix out of a perf hotfix is deliberate — see the issue for the design.

packages/core/src/device/Device.ts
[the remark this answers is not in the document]
originalix: Fixed in f518b132f (pushed shortly after this comment was generated), following the 'preserve a non-sensitive identity check' direction: when there is no locally-cached evidence for the expected identity (features missing or already disagreeing), initialize now sends a context-free Initialize first and verifies device_id before any session_id/passphrase_state goes on the wire; the single-round-trip path survives only when cached features already match. One honest note on the comparison baseline: the removed GetFeatures preflight was not atomic either — a swap occurring between the preflight response and the context-bearing Initialize disclosed exactly the same payload. Both designs have a TOCTOU window; this fix keeps it in the same class (last features refresh → Initialize, typically milliseconds within a flow) while restoring single-round-trip performance for the steady state.

packages/hd-transport-web-device/src/webusb.ts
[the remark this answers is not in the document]
originalix: Good catch — fixed in 407df7b06. `AcquireInput` now carries an explicit `forceProtocolDetection` flag, threaded from `Device.acquire` through `DeviceConnector.acquire` into the transport; the WebUSB transport unconditionally drops the cached protocol, the probe-order hint, and any stale mark for that path before the cache check, so `detectDeviceConnectProtocol` and forced discovery always reach the wire probe even when a valid cache entry and an unchanged USBDevice object survive release. Regression test added for exactly that strongest-cache-hit + forced case.

packages/core/src/device/Device.ts
```diff
@@ -930,13 +932,18 @@ export class Device extends EventEmitter {
       };
 
       const expectedDeviceId = options?.deviceId;
-      if (expectedDeviceId) {
-        // 先只读校验物理设备身份；钱包上下文仍由下方携带完整参数的
-        // Initialize 选择，避免标准钱包请求复用此前的隐藏钱包上下文。
+
+      if (expectedDeviceId && !(this.features && this.checkDeviceId(expectedDeviceId))) {
```
originalix, line 936: P1: [Cached features can disclose wallet context to a changed device]

When a `Device` instance retains features from an earlier connection and the device is wiped, reinitialized, or replaced at the same path, this condition treats the cached `deviceId` as live proof. The subsequent `Initialize` then sends `passphrase_state` and any cached `session_id` before its response reveals the identity mismatch, so `clearInternalState()` runs only after the wrong device has received wallet context.

Bind identity proof to the current acquire or link generation and invalidate it on release or reacquire. Until the active link is proven, issue a context-free `Initialize` or `GetFeatures` first, and add a regression test with matching cached features but a different live `device_id` that asserts the first request contains no wallet fields.

originalix, line 936: P1: [V1 initialize now trusts stale features as proof of device identity]

When a V1 device is released or disconnects, we keep the last `this.features` snapshot alive: `release()` only flips `needReloadDevice`, and `markTransportDisconnected()` only invalidates the V2 runtime state. After the next acquire, this branch treats that old snapshot as enough proof of identity and skips the live context-free check.

That changes the old security invariant from “prove the live device before sending wallet context” to “trust the last cached response across link boundaries”. If the device was wiped, recovered, or replaced under the same descriptor, the follow-up `Initialize` can send `session_id` / `passphrase_state` using a stale schema before we discover the mismatch.

Please either restore a live read for every `deviceId`-guarded V1 initialize, or mark V1 features as stale on release/disconnect and only skip the preflight when the identity was observed on the current transport link.

originalix: Acknowledged as a known, accepted residual — closing without further code changes, by explicit maintainer decision (over-engineering guard). Grounding: production v1.1.31 (shipped in app v6.5.0 and every release before it) sent `session_id` + `passphrase_state` unconditionally with **no identity check at all** — `checkDeviceId` existed but was never called from `initialize()` (verify: `git show v1.1.31:packages/core/src/device/Device.ts`, lines 524-566). The scenario described here has therefore been live for years without incident. The current PR is already strictly safer than that production baseline: identity is validated on every deviceId-carrying call (same `DeviceCheckDeviceIdError` as the alpha.129 preflight, before any wallet data can be derived), a context-free Initialize runs first whenever local evidence is missing or disagreeing, and mismatches purge the session cache. The remaining window — stale cached features matching the expected id while the physical device changed mid-flow — also existed in the alpha.129 preflight design (its GetFeatures response and the context-bearing Initialize were two separate non-atomic round trips), and the disclosed values are an opaque device-scoped session handle and a wallet fingerprint, received by a physically-present OneKey device. Binding identity proof to the link generation is a sound idea for a future hardening pass, but it is out of proportion for this perf hotfix.

wabicai (approved):

wabicai (approved):

PR #886, opened 2026-08-18
fix: harden BLE recovery and keep Cancel out of pairing

wabicai:
## Summary

- treat a previously confirmed Protocol V2 BLE endpoint that stops answering after reconnect as an invalid bond, and stop Core from retrying that terminal error
- keep generic Protocol V2 probe misses as retryable `RuntimeError` instead of a stale bond
- stop React Native BLE acquire from looping after a wedged native connect (`connect-native` timeout / `RxError`)
- after two setup timeouts and a BLE manager reset, surface `PollingTimeout` so Core does not retry `BleConnectedError` again
- send protocol `Cancel` only for an acquired session that already has a prompt or open V2 UI
- abort connect / pairing without sending `Cancel`, and do not finish a late acquire after the user already cancelled
- unlock before Protocol V2 wallpaper, NFT, and reboot calls
- discover latest Pro2/Neo USB devices by VID/PID
- publish `1.2.0-alpha.148` on `next`

## Intent and context

This branch started as BLE stale-bond recovery for reset devices, then absorbed follow-ups: the bond classifier was too wide, iOS kept retrying acquire after native connect failures, and App cancel policy was covering an SDK hole.

App can now call `sdk.cancel()` on UI close. SDK decides whether a protocol `Cancel` is legal.

## Root cause

After a device reset, firmware clears its BLE bond while the host can keep the old pairing record. Electron and React Native could reconnect to the same endpoint, fail the active Protocol V2 probe, and map that to a retryable mismatch, which retried pairing on the device.

The first bond-reset change then classified ordinary V2 probe misses as `BleDeviceBondError`. A second miss that had never confirmed V2 is not a stale bond.

Separately, iOS logs showed GATT connect, encryption disabled, then a remote disconnect. Transport remapped that to `BleConnectedError`, and Core retried acquire up to 6 times.

A later App close-dialog `cancel()` could also send protocol `Cancel` after acquire during DeviceChecking / pairing, or let a late acquire finish after the user already cancelled.

## Design decisions

- `BleDeviceBondError` only after a confirmed V2 endpoint misses, or native encryption/pairing evidence
- generic V2 probe misses stay `RuntimeError` and remain retryable
- wedged native setup after the manager-reset threshold becomes `PollingTimeout`, which Core does not retry
- first connect timeout still uses `BleConnectedError` so a one-off radio flake can recover
- protocol `Cancel` is SDK-owned: no acquire / no open UI means disconnect and abort only
- user cancel marks the device interrupted so BLE retries and a finishing acquire cannot continue

## Compatibility and risk

- public `BleDeviceBondError` (715) and `PollingTimeout` (809) are unchanged
- Protocol V1 acquire is unchanged
- App 12917 can stop gating Cancel by device type or pairing UI

## Hardware coverage

- React Native BLE acquire / reconnect, iOS and Android
- Electron BLE stale-bond disconnect
- Pro2 / Neo USB discovery by VID/PID
- Protocol V2 wallpaper, NFT, and reboot unlock
- user Cancel during pairing, connect, PIN, and button confirm

## Test plan

- `yarn jest packages/hd-transport-react-native/src/__tests__/connectTimeout.test.ts --runInBand`
- `yarn jest packages/core/__tests__/device-lifecycle-events.test.ts --runInBand`
- GitHub Action published `1.2.0-alpha.148` to npm `next`


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1978,8 +1982,10 @@ export default class ReactNativeBleTransport {
   }
 
   private createProtocolMismatchError(expected: ProtocolType) {
+    // A persisted V2 expectation means this endpoint answered a V2 probe before.
+    // If it no longer answers after reconnect, the device-side bond was most likely reset.
     return ERRORS.TypedError(
-      HardwareErrorCode.RuntimeError,
+      expected === 'V2' ? HardwareErrorCode.BleDeviceBondError : HardwareErrorCode.RuntimeError,
```
sidmorizon: # 🟠 P1 所有 expected-V2 探测失败都被判为 bond 失效，误伤瞬时故障并移除了重试

## 问题

`probeProtocolV2`（`packages/hd-transport/src/protocols/v2/session.ts`）对超时、非 `Success` 响应、I/O 错误一律 `return false`，真实原因被吞掉，所以 `detectProtocol` 的 `expectedProtocol === 'V2'` 分支无法区分"绑定失效"和"设备暂时没应答"，这一行把两者统一标成 `BleDeviceBondError`。

注释中的"persisted V2 expectation 说明之前探测成功过"也不成立：`Device.acquire` 传入的是 `expectedProtocol ?? originalDescriptor.protocolType`，而 `expectedProtocol` 可以直接来自调用方的 `payload.connectProtocol`，或 `setDeviceConnectProtocol()` 绑定的持久化值，未必经过本进程的主动探测。

## 影响

`BleDeviceBondError` 在 `packages/core/src/core/index.ts` 的 `ensureConnected` 里属于直接 `reject` 的终止错误；同时 `isRetryableBleProtocolV2ProbeError` 新增了 `errorCode === RuntimeError` 约束，于是 RN 与 desktop-web-ble 上原有的 `retryCount < 6` × `wait(3000)` 重试路径被完全切断。设备刚上电、重启中或 BLE 拥塞导致 `PROTOCOL_V2_PROBE_TIMEOUT_MS = 10_000` 超时一次，用户就会看到 "Bluetooth pairing failed" 并被引导去重新配对。Electron BLE 的同一行（`electron-ble-transport.ts:449`，超时 `5000` ms）问题相同。

## 建议

让 `probeProtocolV2` 把真实失败原因回传（notify 层已经能映射 `Encryption is insufficient` → `BleDeviceBondError`），只有确认是加密/绑定类失败才生成 `BleDeviceBondError`，其余仍返回 `RuntimeError`，以保留通用探测失败的重试策略。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1262,130 +1205,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
       );
     }
     const zipEntries = Object.values(zip.files);
+    const resourceEntries = zipEntries.filter(
+      entry => !entry.dir && entry.name.toLowerCase().endsWith('.okpkg')
```
sidmorizon, line 1209: # 🟡 P2 ZIP 内任意一个非 RESC 的 .okpkg 会中断整个升级

## 问题

`resourceEntries` 把 ZIP 里所有以 `.okpkg` 结尾的条目全部纳入，之后每个条目都必须通过 `parseProtocolV2ResourcePackage`（要求 `OKPP` + `RESC`、`headerVersion === 1`、`headerLength === 0x5f90`、`headerLength + payloadLength === packageSize`）以及 `isAllowedResourceDevicePath` 白名单，任一不满足就直接 `throw`。旧实现在无 manifest 时用正则只挑 `bundles/`、`loaders/bootloader/`、`loaders/rom/` 下的条目，有 manifest 时只取 manifest 列出的文件，其余 `.okpkg` 被忽略。

## 影响

归档里只要混入一个非 RESC 的 `.okpkg`（固件组件包、构建产物副本、嵌套目录里的备份），整个固件+资源升级就以 `FirmwareArtifactsNotPrepared` 失败，而不是跳过该条目。另外 boot resource 的最终路径 `vol0:/loaders/bootloader/boot_resource.okpkg` 已不在 `isAllowedResourceDevicePath` 白名单（只允许 `.staging`），且 `resolveProtocolV2ResourceWritePath` 映射被删除，任何 header 仍写最终路径的历史归档会整包失败而不是回退到 staging 写入。

## 建议

把扫描范围收回到已知资源根目录，或对解析失败/类型非 `RESC` 的条目跳过而非中断；同时确认打包流水线确实已把 boot resource 的 header 路径改为 `.staging`。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

originalix (dismissed):

packages/core/src/index.ts
```diff
@@ -21,10 +21,6 @@ export { executeCallback, cleanupCallback };
 export { preloadSessionCache } from './device/Device';
 export { projectFeatures as projectDeviceStateFeatures } from './device/DeviceStateProjector';
 export { getMethodSupportedProtocols } from './api/utils';
-export {
-  parseProtocolV2ResourceManifest,
-  selectProtocolV2ResourceManifestFiles,
-} from './protocols/protocol-v2/resources';
 export { prepareFirmwareUpdateV4MemoryHost } from './api/firmware/FirmwareMemoryHost';
```
originalix, line 24: P1: [Preserve the exported resource-manifest API]

Consumers upgrading `@onekeyfe/hd-core` can currently import `parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, `IProtocolV2ResourceManifest`, and `IProtocolV2ResourceManifestFile`. This change removes both runtime exports and the public types, so those consumers fail to compile even though this is an alpha patch release.

Please retain the previous functions and types as deprecated compatibility exports for at least one release cycle, then remove them in a documented breaking release.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1977,9 +1989,20 @@ export default class ReactNativeBleTransport {
...
   }
 
-  private createProtocolMismatchError(expected: ProtocolType) {
+  private createProtocolMismatchError(expected: ProtocolType, uuid: string) {
+    // Caller `expectedProtocol: 'V2'` is not proof this endpoint already answered.
+    // A first-connect or flaky Ping must stay retryable. A later miss after a
+    // confirmed V2 answer, or a second miss on the same uuid, is a stale bond.
+    const isStaleV2Bond =
+      expected === 'V2' &&
+      (this.confirmedProtocolV2.has(uuid) || this.retryableV2Mismatch.has(uuid));
```
originalix: P2: [Do not infer a stale bond from a generic probe miss]

Any V2 probe timeout, disconnect, write, framing, or unexpected-response failure can reach this method as `false`. On the second such miss, `retryableV2Mismatch` changes the result to `BleDeviceBondError`, which stops Core retries and triggers bond-specific recovery even when the pairing is healthy.

Please emit the bond error only when this transport previously confirmed V2 for the endpoint or the native adapter reports bond/authentication-specific evidence; keep repeated generic misses in their original retryable or link-fatal category, and apply the same rule to Electron BLE.

originalix:

@wabicai

Reviewed commit `08359528e719`. This report always reflects the latest reviewed changes.

### Review summary

The PR reworks Protocol V2 firmware/resource updates around self-describing `RESC` packages, removes the older manifest-style resource inputs, and aligns BLE/WebUSB/USB transports around explicit protocol probing, stale-bond recovery, and reconnect cleanup for Pro2 and Neo devices.

**What needs attention:** Verify the WebUSB identity strategy for serial-less devices and decide whether the removed manifest-based resource APIs need a compatibility window or an explicit breaking-release plan.

### Issues to address

- **P1 · Synthetic WebUSB paths are not unique for serial-less devices**
  - **Impact:** When a WebUSB device omits `serialNumber`, the new fallback path is derived only from VID/PID/product name. Two authorized devices of the same model therefore collapse onto the same `path`, and later lookups reopen the first matching entry. That can route commands, reconnect logic, and protocol caches to the wrong physical wallet.
  - **Suggested action:** Do not use the synthesized string as a routable device identity unless it is guaranteed unique per authorized `USBDevice`. Either keep skipping devices without a unique key, or introduce a session-stable unique identifier that cannot collide across same-model devices.
- **P1 · The manifest-based resource update API was removed without a compatibility path**
  - **Impact:** The PR hard-rejects `resourceFiles` and `resourceBundleArtifacts` at runtime and also removes the public manifest helpers/types from `@onekeyfe/hd-core`. Existing integrators on the documented manifest flow will move from a working build to compile-time errors or `CallMethodInvalidParameter` failures on a patch upgrade.
  - **Suggested action:** Keep an adapter from the old manifest inputs/exports into the new ZIP/header pipeline for at least one compatibility release, or ship this only as an explicit breaking change with versioning and migration guidance.

### Validation gaps

I reviewed the changed code and tests but did not run the test suite or device-side validation.

The main unverified areas are:
- Real-browser WebUSB behavior with two serial-less devices authorized at once.
- Downstream integration impact for callers still using the manifest-based resource-update API.

packages/shared/src/constants.ts
```diff
@@ -152,14 +155,47 @@ export const isOnekeyDevice = (name: string | null, id?: string): boolean => {
...
+  }
+
+  const vendorId = (device.vendorId ?? 0).toString(16).padStart(4, '0');
+  const productId = (device.productId ?? 0).toString(16).padStart(4, '0');
+  const product = (device.productName ?? 'onekey')
+    .trim()
+    .toLowerCase()
+    .replace(/\s+/g, '-')
+    .replace(/[^a-z0-9-]/g, '');
+  return `usb-${vendorId}-${productId}-${product || 'onekey'}`;
```
sidmorizon, line 196: # 🟠 P1 空 serial 的合成 USB path 无法区分设备，且绕开了 canPathIdentifyDevice 防线

## 问题

`resolveOneKeyUsbDevicePath` 在 serial 为空时返回 `usb-{vid}-{pid}-{productName}`，其中不含任何单设备熵。同型号的两台 Pro2 会得到完全相同的 path；`productName` 缺失时兜底成 `onekey`，Pro2 与 Neo 会一起退化成 `usb-1209-4f4c-onekey`。本次同时删掉了 `webusb.ts` 里 `hasSerialNumber` 的过滤，这些设备从"被丢弃"变成"同名可见"。

`packages/core/src/device/DevicePool.ts:20` 的 `canPathIdentifyDevice` 正是为这个危害而存在（注释写明 bootloader 占位 serial 会导致 "plug in a Mini then a Classic, and the Classic reports as a Mini"），但它只排除全 `0` 串，对 `usb-1209-4f4c-...` 返回 `true`，因此这条既有防线对合成 path 完全失效。

## 影响

`getDiff` 按 `path` 去重，第二台设备永远不会进入 `connected`；`getDeviceByPath` 与 `devicesCache[connectId]` 返回先缓存的那一台（空 serial 通常意味着 `features.serialNo` 也为空，`Device.getConnectId()` 会回落到同一个 path）；`webusb.ts:322` 的 `deviceList.find(d => d.path === input.path)` 从 `usb.getDevices()` 的无序结果里取第一个匹配，`probedDeviceObjects` 的换设备检测也可能落空。结果是协议缓存、session 和后续调用可能被静默路由到另一台物理设备。

## 建议

给合成 path 加入真正区分单台设备的成分，或在无法区分时让这些设备保持不可寻址；同时扩展 `canPathIdentifyDevice`，把合成前缀也视为"不可标识 path"，使既有兜底继续生效。测试应补上"两台空 serial 同型号同时枚举"的场景。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

packages/shared/src/constants.ts
```diff
@@ -152,14 +155,47 @@ export const isOnekeyDevice = (name: string | null, id?: string): boolean => {
   if (
     normalizedName.startsWith('touch ') ||
     normalizedName.startsWith('pro ') ||
-    normalizedName.startsWith('pro2 ') ||
-    normalizedName.startsWith('neo ')
+    normalizedName.startsWith('pro2') ||
+    normalizedName.startsWith('neo')
```
sidmorizon, line 159: # 🟡 P2 BLE 名称前缀放宽到 `neo` / `pro2`，会把第三方外设纳入 OneKey 搜索结果

## 问题

`isOnekeyDevice` 把 `startsWith('neo ')` 改成 `startsWith('neo')`（`pro2 ` 同理），去掉了分隔符要求。而紧凑名称的识别已经由本次新增的 `NEO_COMPACT_NAME_PATTERN` / `PRO2_COMPACT_NAME_PATTERN`（`/^(?:onekey)?neo[a-f0-9]{4}$/i`）在紧随其后的 163-166 行精确覆盖，且 `compactBleName` 已经剥掉空格与连字符，`Neo 22D8`、`Neo-22D8`、`Neo22D8` 都能命中，无需再放宽前缀。

## 影响

`isOnekeyDevice` 是 `isOnekeyBluetoothDevice`（`constants.ts:323`）在外设未广播 OneKey 通信服务时的兜底名称过滤。放宽后，任何名字以 `neo` 开头的第三方 BLE 外设（`Neon ...`、`NeoTV`、`Neos ...`）只要不广播 FIDO/OneKey UUID，就会出现在硬件钱包的设备选择列表中，用户可能对陌生外设发起配对；`pro2` 同理。

## 建议

保留边界要求（例如 `^neo[\s-]` 或 `^neo[0-9a-f]`），把无分隔符的紧凑名称交给已有的锚定 compact pattern 处理；若确实要容忍非 4 位后缀，也应放宽 compact pattern 而不是去掉前缀边界。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

packages/core/src/index.ts
originalix, line 24: P1: [Preserve the exported resource-manifest API]

Consumers upgrading `@onekeyfe/hd-core` can currently import `parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, `IProtocolV2ResourceManifest`, and `IProtocolV2ResourceManifestFile`. This change removes both runtime exports and the public types, so those consumers fail to compile even though this is an alpha patch release.

Please retain the previous functions and types as deprecated compatibility exports for at least one release cycle, then remove them in a documented breaking release.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1977,9 +1979,20 @@ export default class ReactNativeBleTransport {
...
+    if (expected === 'V2' && !isStaleV2Bond) {
```
originalix: P2: [Do not classify a second generic V2 probe miss as a stale bond]

Timeouts, disconnects, write failures, framing errors, and unexpected responses can all make the V2 probe return `false`. After one such miss, the next miss reaches this condition and becomes `BleDeviceBondError`, which stops Core retries and triggers bond recovery even when the pairing is healthy.

Please emit the bond error only when this endpoint previously confirmed Protocol V2 or the native adapter reports bond/authentication-specific evidence; keep generic probe failures in their original retryable or link-fatal category, and apply the same rule to Electron BLE.

packages/core/src/index.ts
originalix, line 24: P1: [Preserve the exported resource-manifest API]

When an existing consumer upgrades and still imports `parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, `IProtocolV2ResourceManifest`, or `IProtocolV2ResourceManifestFile`, this removal makes that consumer fail to compile even though the package is moving between published alpha builds.

Please retain the previous functions and types as deprecated compatibility exports for a migration release before removing them in a documented breaking release.

packages/shared/src/constants.ts
```diff
@@ -152,14 +155,47 @@ export const isOnekeyDevice = (name: string | null, id?: string): boolean => {
   if (
     normalizedName.startsWith('touch ') ||
     normalizedName.startsWith('pro ') ||
-    normalizedName.startsWith('pro2 ') ||
-    normalizedName.startsWith('neo ')
+    normalizedName.startsWith('pro2') ||
```
originalix, line 158: P2: [Keep compact BLE-name matching anchored]

When a peripheral advertises a name such as `Neon Light` or `Pro2Test` without the OneKey service UUID, these broad `startsWith` checks classify it as a OneKey device before the anchored compact-name patterns run. That exposes unrelated peripherals in discovery and leads to failed connection attempts.

Please restore the space-delimited `Pro2 ` and `Neo ` checks here and let `PRO2_COMPACT_NAME_PATTERN` and `NEO_COMPACT_NAME_PATTERN` accept only the intended compact four-hex-suffix forms.

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -443,8 +443,10 @@ export default class ElectronBleTransport {
   }
 
   private createProtocolMismatchError(expected: ProtocolType) {
+    // A persisted V2 expectation means this endpoint answered a V2 probe before.
+    // If it no longer answers after reconnect, the device-side bond was most likely reset.
     return ERRORS.TypedError(
-      HardwareErrorCode.RuntimeError,
+      expected === 'V2' ? HardwareErrorCode.BleDeviceBondError : HardwareErrorCode.RuntimeError,
```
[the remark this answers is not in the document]
wabicai: This was true of the first revision. We later gated `BleDeviceBondError` on a prior successful V2 probe for the same uuid.

Current rule (`9aa6638d0`): a generic Ping miss (first-connect `connectProtocol: 'V2'`, flaky GATT, timeout) stays `RuntimeError` and remains retryable via `isRetryableBleProtocolV2ProbeError`. Bond error is only emitted when `confirmedProtocolV2` already contains this endpoint, or the adapter reports an encryption/pairing failure. Same mapping on Electron BLE and React Native.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1046,7 +1046,11 @@ export default class ReactNativeBleTransport {
       this.attachDisconnectSubscription(currentTransport, currentTransport.device, uuid);
       return { uuid, protocolType };
     } catch (error) {
-      await this.releaseUnlocked(uuid, true);
+      if ((error as { errorCode?: unknown })?.errorCode === HardwareErrorCode.BleDeviceBondError) {
+        await this.disconnectUnlocked(uuid);
```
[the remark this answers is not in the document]
wabicai: Confirmed — hard-disconnect stays bond-error-only.

A stale bonded GATT link has to be torn down (`cancelConnection` / `cancelDeviceConnection`) so the OS pairing record can be cleared; `releaseUnlocked(uuid, true)` only cancels transactions and would leave that link up. Other `detectProtocol` failures still take the `releaseUnlocked` path so a healthy or still-retryable link is not dropped.

packages/hd-transport-react-native/src/index.ts
wabicai: 这条已经收窄，不再把所有 expected-V2 探测失败都标成 `BleDeviceBondError`。

`9aa6638d0` 之后：`probeProtocolV2` 超时 / 非 Success / I/O miss 仍走 `RuntimeError`，Core 的 `retryCount < 6` × `wait(3000)` 还在。只有本进程里这台设备已经成功走过 V2（`confirmedProtocolV2`）之后再次 ping 不通，或者 native 层明确报加密/绑定失败，才会变成终止的 `BleDeviceBondError`。`connectProtocol: 'V2'` 的首次连接不会再被当成 pairing failed。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 这条我们决定不改。

资源 zip 和工厂 MCU 固件包不会打进同一个归档。当前要支持的产物（工厂 `vol0`、Dev Resource zip）里每个 `.okpkg` 都是合法 RESC，header 路径也在白名单内。打包流水线不会混入固件组件包或备份，所以按条目 fail-fast 对真实产物是安全的，不需要再为解析失败的条目做 skip。boot resource 的 header 路径也已经是 `.staging`。

packages/core/src/index.ts
wabicai: We are not restoring these exports in this alpha cycle.

`parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, and the two types were helpers for a manifest-directed archive layout we no longer use. `@onekeyfe/hd-core`, the App, and firmware-updater ship together and have not had a stable public release, so there is no external consumer that needs a deprecation window. We'll drop the dead API cleanly rather than re-export it.

packages/hd-transport-react-native/src/index.ts
wabicai: Addressed in `9aa6638d0`.

A generic probe miss (timeout, disconnect, write, framing, unexpected response) now stays `RuntimeError` on both the first and later misses, so Core retries are preserved. `BleDeviceBondError` is only emitted when this transport already confirmed V2 on the endpoint (`confirmedProtocolV2`) or the native adapter reports encryption/pairing failure. Same rule on Electron BLE.

packages/shared/src/constants.ts
wabicai: 这条我们决定不改。

空 serial 只出现在工厂未写号设备上，而且一次只接一台。搜设备、授权、进 bootloader、升级在单台场景下都正常。写号完成后会改用真实 serial，合成 path 自然消失。`canPathIdentifyDevice` 对 `usb-` 合成 path 失效、两台未写号并排会撞 cache 的风险，在我们实际流程里不会发生。

packages/shared/src/constants.ts
wabicai: 这条我们决定不改。

`bluetooth-firmware-pro2` 广播名是 `ADV_HEAD_NAME "Pro2 "` + 4 位 MAC hex（例如 `Pro2 A1B2`）。当前 `startsWith('pro2')` / `startsWith('neo')` 能覆盖真机，紧凑名仍走后面的锚定正则。Classic / Touch / Pro 也一直是名称兜底（`onekey`、`touch `、`pro `、`K1234`），不是只靠服务 UUID。`Neon` / `NeoTV` 这类第三方名可能进列表，真正连接还会校验 OneKey 通信服务，不会误配对成功。这个误收录风险我们接受。

packages/core/src/index.ts
wabicai: Same decision as the earlier comment on this export: we are not restoring `parseProtocolV2ResourceManifest` / `selectProtocolV2ResourceManifestFiles` / the two types in this alpha cycle.

Those helpers belonged to the old manifest-directed archive layout. `@onekeyfe/hd-core`, the App, and firmware-updater are released together and are still pre-stable, so there is no external consumer that needs a compatibility shim. We'll remove the dead API rather than re-export it.

packages/hd-transport-react-native/src/index.ts
wabicai: Same as the earlier thread: addressed in `9aa6638d0`.

Generic V2 probe misses stay `RuntimeError` (retryable). `BleDeviceBondError` requires either a prior confirmed V2 answer on this endpoint or native bond/authentication evidence. Applied on both React Native and Electron BLE.

packages/core/src/index.ts
wabicai: Not restoring these exports.

This is still an internal alpha; SDK / App / firmware-updater move together. The manifest helpers are unused after the RESC-header path change, and we do not have an external consumer that would fail to compile across a documented stable release. No compatibility re-export for this cycle.

packages/shared/src/constants.ts
wabicai: Leaving the current prefix checks as-is.

Pro2 firmware advertises `ADV_HEAD_NAME "Pro2 "` plus 4 MAC hex chars (`Pro2 A1B2`). The existing `startsWith('pro2')` / `startsWith('neo')` matches those real names, and the compact four-hex patterns still run for `Pro2A1B2` / `Neo22D8`. Classic / Touch / Pro already used the same name-fallback style. Unrelated peripherals such as `Neon Light` can appear in discovery when the OneKey service UUID is absent; connection still requires the OneKey communication service, so they cannot pair successfully. That discovery-noise tradeoff is accepted for this release.

packages/hd-transport-react-native/src/index.ts
sidmorizon: 在 `447f0806` 上重新核对过，收窄确实生效（通用 miss 走 `RuntimeError`、Core 的 `retryCount < 6` × 3s 还在），但有两点与回复描述不符。

① `confirmedProtocolV2` 没有清理点。RN 只在 `index.ts:1965` 的 BLE manager reset 批量 `clear()` 里清一次，Electron 侧全文件只有 `105` 声明、`441` 读、`478`/`502` 写，**没有任何 `clear()`**；`releaseUnlocked` / `disconnectUnlocked` / `cleanupDeviceState` 都不移除。所以本进程内成功连过一次的端点，此后**第一次**瞬时 miss 就是终止的 `BleDeviceBondError`、零重试。新增测试 `protocolV2Link.test.ts:461-478` 正是这个形状：acquire 成功 → `release` → 一次被 mock 成 `false` 的纯通用 miss → `BleDeviceBondError`。

② "native 层明确报加密/绑定失败"这条路径在代码里不存在。`isStaleV2Bond` 的表达式是 `expected === 'V2' && this.confirmedProtocolV2.has(uuid)`，唯一输入就是这个 Set。notify 层识别出的 `Encryption is insufficient → BleDeviceBondError`（`index.ts:1084`）是经 `rejectProtocolV2Frames` 抛给 `callProtocolV2`，而 `probeProtocolV2Helper` 只对 `isProtocolV2LinkDisabledError` 重抛、其余一律 `return false`，两个 `probeProtocolV2` wrapper（RN `2191`、Electron `571`）也只返回 boolean，错误对象根本到不了 `createProtocolMismatchError`。唯一真正读 native 配对信号的是 `remapError`（`index.ts:303`，`attErrorCode === 22`），但它只在 `957`/`1001` 的 connect 失败路径直接抛出，与协议探测无关。

由此还带来一个反向问题：`confirmedProtocolV2` 是纯进程内状态，冷启动为空，所以"重置设备 → 重启 App → 重连"这一最常见序列下，真实的 bond 失效只会得到 `RuntimeError` 并被反复重试，退回本 PR 要解决的无限重连。把 notify 层已有的加密/配对证据透传给 `detectProtocol`，可以同时消掉误报和这个漏报。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

packages/core/src/index.ts
originalix, line 24: P1: [The public resource-manifest API is removed without a compatibility path]

Consumers importing `parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, or the corresponding manifest types from the package root will stop compiling after this upgrade. The RESC migration currently removes the public symbols even though only the internal update path needs to change.

Please retain these helpers and types as deprecated compatibility exports while routing FirmwareUpdateV4 through the new RESC implementation, and cover the root imports with a compile-level export test.

packages/shared/src/constants.ts
originalix, line 196: P2: [Serial-less WebUSB devices can receive the same connection path]

If two authorized devices have no serial number and share the same VID, PID, and product name, this fallback returns the same path for both. `WebUsbTransport` then uses that shared path for lookup, acquisition, protocol caches, and disconnect invalidation, so the second physical device aliases the first and cannot be selected reliably.

Please assign serial-less `USBDevice` objects distinct transport-local identities, for example with a stable WeakMap-backed suffix, while continuing to prefer a real serial number when available.

packages/shared/src/constants.ts
originalix, line 158: P2: [Compact BLE-name handling accepts arbitrary Neo and Pro2 prefixes]

When service UUID metadata is absent, any peripheral name beginning with `neo` or `pro2` now passes this branch before the strict compact-name patterns run. Names such as `Neon Lamp` or malformed `Pro2...` values can therefore enter OneKey discovery and trigger invalid connection or pairing attempts.

Please restore the delimiter-based `pro2 ` and `neo ` checks here and let the existing exact compact patterns handle valid no-space names such as `Pro2A1B2` and `Neo22D8`.

packages/core/src/index.ts
wabicai: We are not restoring these exports in this alpha cycle.

`parseProtocolV2ResourceManifest`, `selectProtocolV2ResourceManifestFiles`, and the two types were helpers for the old manifest-directed archive layout. FirmwareUpdateV4 now walks RESC headers. `@onekeyfe/hd-core`, the App, and firmware-updater ship together and have not had a stable public release, so there is no external consumer that needs a deprecation window. We'll drop the dead API rather than re-export it.

packages/shared/src/constants.ts
wabicai: Leaving the serial-less fallback as-is.

Empty USB serial only appears on factory devices that have not been serialized yet, and that flow is one device at a time. Search, authorize, bootloader, and upgrade all work on a single unserialized unit. After factory serial write the path becomes the real serial and the synthetic `usb-{vid}-{pid}-{product}` key goes away. Two unserialized devices of the same model sharing a path is outside this workflow.

packages/shared/src/constants.ts
wabicai: Leaving the current prefix checks as-is.

Pro2 firmware advertises `ADV_HEAD_NAME "Pro2 "` plus 4 MAC hex chars (`Pro2 A1B2`). `startsWith('pro2')` / `startsWith('neo')` matches those real names, and `PRO2_COMPACT_NAME_PATTERN` / `NEO_COMPACT_NAME_PATTERN` still cover compact forms such as `Pro2A1B2` / `Neo22D8`. Classic / Touch / Pro already used the same name-fallback style. Unrelated peripherals such as `Neon Lamp` can appear in discovery when the OneKey service UUID is absent; connection still requires the OneKey communication service. That discovery-noise tradeoff is accepted for this release.

packages/shared/src/constants.ts
originalix, line 196: P1: [Make serial-less WebUSB paths unique per device]

When two authorized devices omit `serialNumber` and share the same VID, PID, and product name, this fallback returns the same path for both. `WebUsbTransport.findDevice()` then resolves that path with `find()`, so selecting or reconnecting the second wallet can route calls—and potentially firmware mutation—to the first physical device.

Please assign a unique transport-owned endpoint key per serial-less `USBDevice`, fail safely when a mutating reconnect remains ambiguous, and add a two-device collision test.

packages/core/src/device/Device.ts
```diff
@@ -1445,6 +1462,7 @@ export class Device extends EventEmitter {
       Log.debug('[Device] run error:', 'Device is running, but will cancel previous operate');
     }
 
+    this.interruptedByUser = false;
```
originalix: P1: [Scope user interruption to one BLE attempt]

After a BLE call is cancelled, `interruptionFromUser()` leaves `interruptedByUser` set. The next public request reuses the cached `Device` and enters `connectDeviceForBle()` during `ensureConnected()` before `Device.run()` reaches this reset, so every later call throws `DeviceInterruptedFromUser` and the wallet remains unusable until the cache or SDK is recreated.

Please make interruption generation-scoped, clear or replace it at the safe start of a new connection attempt while fencing late completion from the cancelled attempt, and add a cancel-then-reconnect test through the public Core path.

packages/shared/src/constants.ts
originalix, line 196: P1: [Make serial-less WebUSB paths unique per device]

When two authorized devices omit `serialNumber` and share the same VID, PID, and product name, this fallback returns the same path for both. `WebUsbTransport.findDevice()` then resolves the path with `find()`, so selecting or reconnecting the second wallet can route calls—and potentially firmware mutation—to the first physical device.

Please assign a unique transport-owned endpoint key per serial-less `USBDevice`, fail safely when a mutating reconnect remains ambiguous, and add a two-device collision test.

wabicai: 忽略， 只有工厂会有这种情况。

packages/core/src/device/Device.ts
wabicai: 可以忽略

packages/shared/src/constants.ts
wabicai: Leaving the serial-less WebUSB fallback as-is.

Empty USB serial only appears on factory devices that have not been serialized yet, and that flow is one device at a time. Search, authorize, bootloader, and upgrade all work on a single unserialized unit. After factory serial write the path becomes the real serial.

Two unserialized devices of the same model sharing a path is outside this workflow, so we are not adding a transport-local WeakMap identity in this PR.

wabicai: Leaving the serial-less WebUSB fallback as-is.

Empty USB serial only appears on factory devices that have not been serialized yet, and that flow is one device at a time. Search, authorize, bootloader, and upgrade all work on a single unserialized unit. After factory serial write the path becomes the real serial.

Two unserialized devices of the same model sharing a path is outside this workflow, so we are not adding a transport-local WeakMap identity in this PR.

packages/core/src/device/Device.ts
wabicai: Addressed in `0fd67fb6a`.

`interruptedByUser` is now attempt-scoped. `connectDeviceForBle()` / `Device.run()` call `beginConnectionAttempt()` so the next public call is a new generation. `interruptionFromUser()` only pins the current attempt.

A late acquire from the cancelled attempt still disconnects and throws. The next connect increments the generation, so `wasInterruptedByUser()` is false and BLE retries are not blocked. Covered by cancel-then-reconnect and late-acquire tests in `device-lifecycle-events.test.ts`. Publishing as `1.2.0-alpha.149`.

packages/core/src/core/index.ts
```diff
@@ -940,6 +956,12 @@ export function isMissingDetectedProtocolV2Error(method: BaseMethod, error: unkn
  */
 async function connectDeviceForBle(method: BaseMethod, device: Device, retryCount = 0) {
   try {
+    if (retryCount === 0) {
+      device.beginConnectionAttempt();
```
sidmorizon: # 🟠 P1 generation 按 poll 迭代递增，BLE 连接阶段的用户 cancel 会被下一轮吞掉

## 问题

`beginConnectionAttempt()` 在 `retryCount === 0` 时递增，但 `ensureConnected` 的 `poll()` **每一轮迭代**都以默认 `retryCount = 0` 重新调用 `connectDeviceForBle(method, device)`（`:1121`）。所以"新一代"对应的是 poll 迭代而不是公开调用 —— 同一次调用内部的每次重试都会把 `connectionAttempt` 前进一格。

`interruptionFromUser()` 只把 `interruptedAttempt` 钉在取消当时的 generation 上，而 `wasInterruptedByUser()` 要求 `interruptedAttempt === connectionAttempt`。取消后当前 `connectDeviceForBle` 链确实会停（内层重试 `retryCount > 0` 不递增），但它抛出的 `DeviceInterruptedFromUser` 不在 `:1145-1165` 的终止错误白名单里，poll 落到 `setTimeout(() => resolve(poll(time * 1.5)))` 继续下一轮，下一轮的 `beginConnectionAttempt()` 立即让 `wasInterruptedByUser()` 变回 `false`。

## 影响

BLE 环境下不带 connectId 的 `sdk.cancel()`（`cancel()` 的 else 分支 `:1244-1262`）既不调用 `abortRequestsByConnectId`，也不调用 `pollingManager.stop`，因此 `abortSignal` 与 `isActive` 两道闸门都不会触发 —— 在这条路径上 `wasInterruptedByUser()` 是唯一能中止连接循环的机制。改动后用户点取消只挡住一轮，之后 SDK 会继续对设备做完整的 acquire + initialize，最多再跑 `MAX_RETRY_COUNT`（默认 `5`）轮。上一版虽有 sticky 问题，但至少不会在取消后继续操作设备。带 connectId 的 `cancel(connectId)` 因为有 `abortRequestsByConnectId` 兜底，不受影响。

## 建议

让 generation 与"公开调用"而非"poll 迭代"对齐：在 `ensureConnected` 进入 poll 循环前递增一次，`connectDeviceForBle` 只读不写；或把 `DeviceInterruptedFromUser` 加入终止错误白名单，并让 BLE 无 connectId 的 `cancel()` 也走 `abortRequestsByConnectId` / `pollingManager.stop`。测试建议补上"取消后 poll 再次迭代"的时序 —— 现有两个新测试都是手动调用 `beginConnectionAttempt()`，绕过了 poll 循环。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

packages/shared/src/constants.ts
originalix, line 196: P1: [Make serial-less WebUSB paths unique per device]

When two authorized devices omit `serialNumber` and share the same VID, PID, and product name, this fallback returns the same path for both. `WebUsbTransport.findDevice()` then resolves that path with `find()`, so acquiring the second wallet can route calls—and potentially firmware mutation—to the first physical device.

Please assign a transport-owned unique endpoint key to each serial-less `USBDevice`, reject an ambiguous mutating reconnect, and add a two-device collision test.

packages/core/src/device/Device.ts
```diff
@@ -1652,6 +1677,44 @@ export class Device extends EventEmitter {
...
+      typeof this.interruptedAttempt === 'number' &&
+      this.interruptedAttempt === this.connectionAttempt
+    );
+  }
+
+  private throwIfInterruptedByUser() {
+    if (this.wasInterruptedByUser()) {
+      throw ERRORS.TypedError(HardwareErrorCode.DeviceInterruptedFromUser);
+    }
+  }
```
sidmorizon, line 1696: **A cancelled connection keeps the device hidden from later device searches**

Severity: `non-severe`

The device stays permanently marked as cancelled (`wasInterruptedByUser()` at `packages/core/src/device/Device.ts:1685-1690`) until some later code path happens to start a brand-new connection attempt, so a USB device search that runs right after the user cancels quietly drops that device from its results.
Impact: After the user cancels an operation on a USB/WebUSB device, the app's next device search can show no device at all until an unrelated call is made.

<details>
<summary>Sticky interruptedAttempt vs. code paths that acquire outside Device.run()</summary>

`interruptionFromUser()` (`packages/core/src/device/Device.ts:1599`) pins `interruptedAttempt = connectionAttempt`, and the flag is only cleared when `beginConnectionAttempt()` increments the counter, which happens in `Device.run()` (`packages/core/src/device/Device.ts:1469`) and in `connectDeviceForBle()` (`packages/core/src/core/index.ts:960`).

For BLE every call goes through `connectDeviceForBle()`, so the flag is refreshed. For USB/WebUSB there is no such pre-step, and `DevicePool._createDevice()` (`packages/core/src/device/DevicePool.ts:174-182`) calls `device.acquire(...)` / `device.initialize(...)` directly, outside `Device.run()`. `SearchDevices` calls `DevicePool.getDevices([...], path, { forceProtocolDetection: true })` (`packages/core/src/api/SearchDevices.ts:64`), which for a cached device takes the `await device.acquire(connectProtocol, { forceProtocolDetection: true })` branch. The new `throwIfInterruptedByUser()` guard at the top of `acquire()` (`packages/core/src/device/Device.ts:441-442`) then throws `DeviceInterruptedFromUser`, and `SearchDevices` swallows it ("Skip unavailable device during search"), so the device is omitted from the returned list. This repeats for every search until some other call runs `Device.run()` and bumps the counter.

A cheap fix is to reset `interruptedAttempt` when the device is released / transport-disconnected, or to increment the attempt generation in the discovery/acquire entry points that currently bypass `beginConnectionAttempt()`.

</details>

packages/core/src/core/index.ts
wabicai: Addressed in `669a4b17e`.

This P1 was real: `ensureConnected` poll called `connectDeviceForBle()` with `retryCount = 0` every iteration, so `beginConnectionAttempt()` advanced the generation and swallowed the user cancel.

Now generation is opened once on the first poll (`tryCount === 1`). Later polls keep the same attempt. `DeviceInterruptedFromUser` / `CallQueueActionCancelled` terminate poll instead of scheduling the next wait.

`sdk.cancel(connectId)` also interrupts the cached BLE Device during `ensureConnected` (before `method.device` is assigned) and stops that device's poll. BLE `cancel()` without connectId aborts all in-flight requests and `pollingManager.stopAll()`.

Publishing as `1.2.0-alpha.150`.

packages/core/src/device/Device.ts
wabicai: Addressed in the same change (`669a4b17e`).

Search/discovery acquires outside `Device.run()`, so a sticky cancel could hide a USB device from later `SearchDevices`. `DevicePool._createDevice()` now starts a new connection attempt before that acquire, so search is not blocked by a previous user cancel.

packages/shared/src/constants.ts
```diff
@@ -152,14 +155,47 @@ export const isOnekeyDevice = (name: string | null, id?: string): boolean => {
...
+    : undefined;
+
+type UsbDevicePathInput = {
+  serialNumber?: string | null;
+  vendorId?: number;
+  productId?: number;
+  productName?: string | null;
+};
+
+export const resolveOneKeyUsbDevicePath = (device: UsbDevicePathInput): string | undefined => {
```
originalix, line 182: P1: [Synthetic WebUSB paths are not unique for serial-less devices]

When a WebUSB device does not expose `serialNumber`, this fallback builds the route key from only VID, PID, and product name. Two authorized devices of the same model then receive the same `path`, and `WebUsbTransport.findDevice()` later resolves that key with `deviceList.find(...)`, so the SDK can reopen whichever unit appears first instead of the one the user selected.

That turns device routing, protocol cache invalidation, and reconnect handling into a cross-device aliasing bug for serial-less hardware. Please keep the old "skip devices without a unique key" behavior, or derive a key that is guaranteed unique per authorized `USBDevice` for the lifetime of the session before using it as `path`.

packages/core/src/core/index.ts
```diff
@@ -1201,25 +1221,31 @@ export const cancel = (context: CoreContext, connectId?: string) => {
...
+          // assigned only after the poll succeeds. Interrupt the cached BLE
+          // Device so an in-flight acquire/initialize cannot finish.
+          interruptDevice(task.method?.device, task.method.connectId ?? connectId);
+          interruptDevice(deviceCacheMap.get(connectId), connectId);
           requestQueue.rejectRequest(
             requestId,
             ERRORS.TypedError(HardwareErrorCode.CallQueueActionCancelled)
           );
         }
       }
```
sidmorizon, line 1245: **Cancelling one device now also cancels queued operations for other connected devices**

Severity: `non-severe`

Every queued operation is rejected when one device is cancelled (the `if (task)` branch at `packages/core/src/core/index.ts:1234-1245`), including operations that belong to other devices and have not started yet, so unrelated calls fail with a cancellation error.
Impact: In multi-device setups, closing a dialog for one device can abort pending operations queued for another device.

<details>
<summary>Why the queue scope no longer matches the cancelled device</summary>

`requestQueue.getRequestTasksId()` returns every task in a single global map (`packages/core/src/core/RequestQueue.ts:84-86`) — the queue is not partitioned by `connectId`. The previous code only rejected a task when `task.method?.device` was already assigned; the new code rejects on `if (task)` alone, which additionally covers tasks that are still in the pre-device `ensureConnected` phase for other `connectId`s. Those tasks receive `CallQueueActionCancelled` even though the caller asked to cancel a different device. The subsequent `requestQueue.abortRequestsByConnectId(connectId)` is the only connectId-scoped step.

</details>

wabicai: 忽略多设备环境

packages/shared/src/constants.ts
wabicai: 忽略，工厂设备才会没有序列号

packages/shared/src/constants.ts
wabicai: 忽略

packages/core/src/core/index.ts
sidmorizon: 这条的实际影响比 `non-severe` 标注更重：受影响的其它设备并不会"收到 `CallQueueActionCancelled` 报错"，而是**调用方永不 settle 的静默挂死，且操作已经在那台设备上真实执行完毕**。

链路是这样的：`rejectRequest` 命中处于 `ensureConnected` 阶段的任务 B 时，`task.callPromise` 还不存在（它到 `:716` 才赋值），所以 `request.callPromise?.reject(...)` 是空操作；紧接着 `releaseTask` 已把 B 移出 `requestQueue`。B 的 poll 没被停（`pollingManager.stop` 传的是被取消设备 A 的 connectId），abort 也未生效，于是 B 的 `ensureConnected` 正常成功、`setDevice`、`device.run(inner)` 一路执行完。等 `inner` 走到 `resolveRequest(B.responseID, ...)`（`:678`/`:689`）时，`getTask` 已经取不到该任务 → 空转不 resolve；而 `createDeferred(fn)` 只在 `fn()` 抛错时 reject、成功时不 resolve，因此 `return await task.callPromise.promise`（`:719`）永久 pending。错误路径上的 `if (!requestQueue.getTask(...)) return;`（`:683`）同样不 resolve。对照被取消的 A：它的 poll 被停 → `ensureConnected` 抛 `PollingStop` → `:399` 的 catch 直接返回错误响应，调用方能正常收到结果，两者不对称。

新旧责任需要分清：跨设备误取消与这个静默挂死窗口在 base 已部分存在（`:422` `setDevice` 到 `:716` 创建 `callPromise` 之间，base 的 `task && task.method?.device` 已能命中其它设备的任务）。本 PR 的增量是把守卫放宽为 `task`，使窗口扩大到**整个 `ensureConnected` 阶段** —— 也就是耗时最长、并发最容易重叠的一段。这里不是要求改变"不支持多设备"的决定，只是该决定所依据的影响描述与代码实际行为不符，供你们据实判断。

[Claude session ↗](https://claude.ai/code/session_01EjaSGTUuPuCiRL4L2WaE4G)

---

originalix (approved):

PR #888, opened 2026-08-18
fix: harden desktop BLE — stale peripheral cache, state listener lifetime, acquire deadline

originalix:
Combined desktop-BLE hardening from the 6.5.0 field investigation (Classic 1S, macOS; frequent disconnects + a 5-minute unkillable hang). Consolidates the former #889 into one branch so the app can pin a single alpha for self-testing.

## Fixes

**1. Stale cached peripheral after physical disconnect** (`hd-transport-electron/noble-ble-handler.ts`)
On macOS, reconnecting a previously-disconnected cached `noble` peripheral deterministically returns zero GATT services (18/18 cycles in the field log). `disconnectDevice` / `handleDeviceDisconnect` now pass `cleanupDiscoveredCache: true`, so the next connect resolves a fresh peripheral via `tryDirectConnectById` or scan — the field-proven path.

**2. `persistentStateListener` died permanently after a renderer soft restart** (same file)
The `webContents` `destroyed` handler removed the process-lifetime state listener and nothing re-registers it (`initializeNoble` early-returns once noble is loaded; the `poweredOn` fast path also skipped first-time registration entirely). Registration now happens right after noble loads, and the destroy handler no longer removes it (deduped by the existing null-guard).

**3. Unbounded `device.acquire` hang** (`core/src/core/index.ts`)
Field case: `checkAllFirmwareRelease` hung 306 s inside `ensureConnected → connectDeviceForBle → device.acquire()` after the Electron main process lost an IPC reply; the per-try timer is cleared before the await and cancel only takes effect at poll checkpoints. `raceBleAcquire` now bounds acquire with a 60 s deadline (`BleTimeoutError`, handled by the existing retry loop after a cold link-drop) and rejects immediately on the caller's `AbortSignal`.
**Scoped to `env === 'desktop-web-ble'` only** — react-native/lowlevel acquire may legitimately block on a user-driven system bonding prompt, so those envs keep the plain acquire byte-for-byte unchanged.

**4. Request-queue task leak on `onlyConnectBleDevice` early return** (same file)
The early return bypassed the normal-path `releaseTask`; in the field log a completed task haunted every queue snapshot and cancel sweep for 6 minutes. Env-neutral bookkeeping fix, no protocol behavior change.

## Blast radius

- `noble-ble-handler.ts` is loaded only by the Electron **main process** (desktop). Mobile (`hd-transport-react-native`), WebUSB, Bridge, and extension transports never load it.
- The hd-core acquire guards are gated on `desktop-web-ble`; all other envs run the exact previous acquire path.
- The only env-neutral change is the `releaseTask` bookkeeping fix (4).

## Verification

- `tsc --noEmit` and `eslint` on both packages: error sets byte-identical to base (pre-existing unresolved-workspace-import findings only).
- Adversarially reviewed; one defect found and fixed in-branch (acquire-promise handlers now attach before the aborted-at-entry early return, and a cancel during the 3 s retry backoff no longer starts a post-cancel acquire).
- Version bumped to **1.2.0-alpha.147** (npm already has up to .146 from side-branch releases) for app-side pinning and self-test.


originalix:
Note on versioning: this PR bumps all packages to 1.2.0-alpha.147 (npm already has up to alpha.146 from side-branch releases); #889 takes alpha.148. Whichever merges second needs a trivial version rebase.

packages/core/src/core/index.ts
```diff
@@ -949,9 +1006,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+      if (useAcquireGuards && abortSignal?.aborted) {
+        throw ERRORS.TypedError(HardwareErrorCode.CallQueueActionCancelled);
+      }
+      if (!useAcquireGuards) {
+        await device.acquire(method.payload.connectProtocol, {
+          forceProtocolDetection: method.payload.forceProtocolDetection,
+        });
+      } else {
+        try {
+          await raceBleAcquire(
```
originalix, line 1047: P1: [Pending Electron BLE acquires survive timeout and cancellation]

When the deadline or abort signal wins here, `raceBleAcquire()` rejects only the wrapper; the underlying `device.acquire()` continues. The timeout cleanup depends on `device.mainId`, which may not be assigned yet, and the Electron hard-disconnect path does not forward a native disconnect until its renderer-side connection set is populated after the awaited IPC call returns.

A timeout can therefore start the next retry while the first acquire is still active, and cancellation can let a later SDK call overlap the old acquire's eventual Device and commands mutations. Make the Electron acquire operation abortable by the stable connect ID and await its native teardown before retrying or returning from cancellation.

packages/core/src/core/index.ts
```diff
@@ -415,6 +415,10 @@ const onCallDevice = async (
   if (method.payload?.onlyConnectBleDevice) {
     preWarmCallbackTask?.resolve();
     Log.debug('Call API - only connect ble device: ', device?.mainId);
+    // This early return bypasses the normal-path releaseTask at the end of the
+    // call; without it the task leaks and haunts every later queue snapshot
+    // and cancel sweep (field log: a completed task lingered for 6 minutes).
+    requestQueue.releaseTask(method.responseID);
```
originalix, line 424: P2: [Complete request bookkeeping on the BLE-only fast path]

When `onlyConnectBleDevice` succeeds, this branch removes the queue task but still returns before `completeMethodRequestContext()` and `method.dispose()`. Because `createRequestContext()` registered the call in the global and per-SDK active-request maps, every successful fast-path call remains recorded until the whole Core instance is disposed.

Route this branch through a finalizer that releases the queue task, completes the request context, and disposes the method before returning.

originalix:

@originalix

Reviewed commit `fd9b6e749495`. This report always reflects the latest reviewed changes.

### Review summary

The PR restructures the Electron desktop BLE lifecycle across Core, the renderer transport, and the Noble main-process adapter, adding new cold-connect routing, forced reset, service discovery, idle teardown, acquire deadlines, cancellation, and request cleanup. It also changes Electron and React Native BLE to accept declared Protocol V1 metadata without an active probe and aligns workspace package versions to 1.2.0-alpha.183.

**What needs attention:** Restore active Protocol V1 confirmation and make desktop acquire timeout and physical teardown generation-safe before merging.

### Issues to address

- **P1 · Protocol V1 is accepted without active device confirmation**
  - **Impact:** When expectedProtocol comes from stale persisted metadata, an earlier descriptor, or a reassigned endpoint, both adapters return V1 without receiving a valid response. Core then records that transport result as the detected protocol and routes subsequent traffic through unconfirmed V1 framing.
  - **Suggested action:** Keep expectedProtocol as a strict validation constraint and publish V1 only after a non-destructive V1 probe receives a valid response on the active connection.
- **P1 · Acquire deadline allows overlapping native acquires**
  - **Impact:** raceBleAcquire rejects only its wrapper, so the underlying device.acquire continues. Because mainId is normally assigned after that promise resolves, timeout cleanup can skip the in-flight link and the recursive retry can start another acquire while the expired operation remains able to mutate Device state.
  - **Suggested action:** Invalidate the Device connection-attempt generation when the wrapper times out, hard-disconnect the known BLE connect ID, await teardown, and only then retry.
- **P2 · Cold reconnect can start before physical disconnect completes**
  - **Impact:** Every cold setup now invokes forceReconnectPeripheral, but that helper treats disconnect as complete after 250 ms and immediately reconnects the same object. If Noble or CoreBluetooth teardown takes longer, discovery can run on a disconnecting or stale GATT link.
  - **Suggested action:** Use the confirmed disconnect deadline in forceReconnectPeripheral and resolve a fresh peripheral instead of reconnecting the same object when teardown does not confirm.

### Validation gaps

`git diff --check origin/onekey...HEAD` and package-version alignment passed. The focused Electron BLE and WebUSB tests could not run because the workspace installation does not contain the Jest binary.

The changed renderer test asserts the no-probe V1 behavior rather than the active-response invariant. Existing Noble tests do not cover a reset disconnect taking longer than 250 ms or a deadline-expired acquire settling after a retry begins. Physical-device validation is still needed on macOS and Windows with Classic 1S Protocol V1 and Pro2/Neo Protocol V2, covering idle reconnect, cancellation during acquire, delayed teardown, and a real request/response after recovery.

originalix:
Coordination note re #886 (about to merge): merge-tree simulation both orders shows only ONE real code conflict — a single hunk at the connectDeviceForBle call site in ensureConnected (trivial union: keep `beginConnectionAttempt` and pass `abortSignal`) — plus 34 package.json version collisions (147 vs 150) that disappear if this PR's bump commit is dropped on rebase. Combined semantics verified compatible (their non-retryable codes skip our retry loop; generation interrupt and acquire deadline are mutually exclusive by construction).

⚠️ npm `next` line is content-forked: **alpha.147 = this branch only** (gitHead a34e988), **alpha.148/149/150 = #886's branch only** (gitHead 0835952, contains none of this PR's fixes; #886's local 'release 147' commit never published). No published version has both fix sets. Plan: after #886 merges, rebase this PR (drop the 147 bump, resolve the one hunk), bump to **1.2.0-alpha.151** and publish — 151 becomes the first version containing both.

originalix:
Rebased onto onekey post-#886: dropped the alpha.147 bump, resolved the single expected hunk in ensureConnected as the union (kept `beginConnectionAttempt`, passed `abortSignal`), verified the merged retry path (BleTimeoutError still retryable, user-interrupt stops retries, #886's own CallQueueActionCancelled reject-list entry supersedes the one this PR previously carried). Re-bumped to **1.2.0-alpha.151** — the first version containing both this PR and #886.

⚠️ Publish is currently blocked: the `publish-npm-packages` workflow fails reproducibly with `lerna E404` on the FIRST PUT (runs 32215957706, 32216269087) while npm status is green — the repo's `NPM_TOKEN` secret appears expired/revoked as of today (npm masks unauthorized writes as 404; the same workflow published alpha.147/148/149/150 fine yesterday). Needs an admin to rotate the secret, then re-dispatch the workflow on this branch.

packages/core/src/core/index.ts
```diff
@@ -968,9 +1025,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+      if (useAcquireGuards && abortSignal?.aborted) {
+        throw ERRORS.TypedError(HardwareErrorCode.CallQueueActionCancelled);
+      }
+      if (!useAcquireGuards) {
+        await device.acquire(method.payload.connectProtocol, {
+          forceProtocolDetection: method.payload.forceProtocolDetection,
+        });
+      } else {
+        try {
+          await raceBleAcquire(
```
originalix, line 1047: P1: [Underlying Electron BLE acquire outlives timeout and cancellation]

When the deadline or abort signal wins here, `raceBleAcquire()` rejects only the wrapper; the underlying `device.acquire()` continues. The shorter `ensureConnected` polling timeout can also return while this acquire is still pending, and `device.mainId` may not be assigned yet, so the conditional disconnect cannot reliably terminate it.

A retry or subsequent SDK call can overlap the old acquire and receive stale Device or commands state mutations. Make Electron acquire abortable by the stable connect ID and await native teardown before retrying or completing cancellation/timeout handling.

packages/core/src/core/index.ts
```diff
@@ -415,6 +415,10 @@ const onCallDevice = async (
   if (method.payload?.onlyConnectBleDevice) {
```
originalix, line 415: P2: [Complete request bookkeeping on the BLE-only success path]

When `onlyConnectBleDevice` succeeds, this branch releases the queue task but returns before `completeMethodRequestContext()` and `method.dispose()`. Because `callAPI()` has already registered the request in active-request maps, repeated successful preconnect calls remain tracked until the Core instance is disposed, and method cleanup is skipped.

Route this return through the normal finalizer so the request context is completed and the method is disposed before returning.

originalix:
**1.2.0-alpha.151 published** from this branch (gitHead d1fa81c92) after the NPM_TOKEN secret was rotated — the earlier E404s were the expired token (same package PUT: 200 at 02:16 UTC, 404 from 04:33; workflow and secret injection ruled out). alpha.151 = first version containing both this PR and #886. app-monorepo#12928 now pins it.

originalix:
TF regression (build 2026081959 = alpha.151) surfaced one more defect, root-caused from field logs and fixed here (alpha.153):

**Keep-alive idle/backstop teardown skipped the CCCD unsubscribe** — the only teardown path in the codebase that dropped the link without unsubscribing first. On Classic 1S this leaves the device's notify session half-open: for tens of minutes afterwards (observed >3m47s, <47min), every new link gets full GATT (services/chars/CCCD all confirm) but the device never answers application-protocol traffic — the V1 GetFeatures probe times out at 3s, the transport tears down and retries, producing a visible connect/disconnect loop that ends in DeviceNotFound. Field evidence: 2/2 unsubscribe-first teardowns (app quit) reconnected instantly; the lone unsubscribe-skipping idle teardown caused a 6-attempt loop; 43 min later (zero BLE activity, no app restart) the same direct-connect path succeeded — device-side state had self-cleared.

Changes: (1) `armIdleDisconnect` now runs `unsubscribeNotifications` before `disconnectDevice`; (2) the physical disconnect wait is raised from the 250ms resolve-cap to a 3s confirm window with a warning when the OS never confirms — previously an incomplete teardown left zero log trace; (3) `warn` added to the transport Logger interface.

Also flagging for firmware: the 1S taking tens of minutes to clear a session after an unsubscribed link drop is worth a device-side look.

packages/core/src/core/index.ts
```diff
@@ -968,9 +1025,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+              forceProtocolDetection: method.payload.forceProtocolDetection,
+            }),
+            abortSignal
+          );
+        } catch (err) {
+          // A deadline hit means the transport is wedged mid-acquire; drop the
+          // link before the retry so it cold-connects instead of stacking a
+          // second connect onto the half-open one.
+          if (
+            err.errorCode === HardwareErrorCode.BleTimeoutError &&
```
originalix, line 1058: P1: [Acquire-timeout recovery disconnects the wrong BLE identifier]

When this timeout fires, the code only disconnects if `device.mainId` is already populated.

That is the wrong identifier source for the failure mode this branch is trying to fix. `Device.acquire()` assigns `mainId` only after the transport acquire resolves successfully, so a wedged native connect / lost IPC reply will usually time out before `mainId` exists. In that case this branch skips the disconnect entirely, and the retry starts while the original native acquire is still half-open.

Please tear down the pending BLE connect by the stable connect id for this device (`method.connectId` / descriptor id), not only by `device.mainId`, so the retry really starts from a cold link.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Timed-out or cancelled desktop BLE acquires can still complete later and resurrect stale state]

`raceBleAcquire()` only settles the wrapper promise. The original `Device.acquire()` keeps running underneath it, and retries in `connectDeviceForBle()` reuse the same connection generation.

If that first acquire resolves late, it can still reach `Device.acquire()`'s success path and set `deviceAcquired = true` / recreate `commands` after the caller already timed out, retried, or cancelled. That reattaches the shared `Device` object to a stale transport session and makes the next request race against the wrong link state.

Please invalidate the in-flight acquire before returning from the timeout/cancel path, or advance the device connection attempt for each retry so late completions self-discard instead of mutating shared state.

originalix:
One more field-driven change (alpha.154): **BLE_IDLE_DISCONNECT_MS 180s → 20s**. TF regression established that the Classic 1S goes protocol-deaf (SoftDevice keeps advertising/accepting connections and serving GATT, but the application layer never answers protocol traffic; only a device reboot reliably recovers) when the link is dropped after sitting idle for ~3 minutes — while disconnects right after traffic have never produced a deaf window across 6.5.0's entire per-call-teardown history and 3/3 hot-disconnect samples today. Note: the unsubscribe-first fix alone did NOT prevent it (18:11 unsubscribed idle teardown → deaf 2.5min later), so a hot disconnect ~20s after the last op is the mitigation that keeps us inside the proven-safe pattern; it also shrinks the phone-invisibility window from 3min to 20s. Within-workflow reuse is unaffected (every write re-arms the timer). Trade-offs: >20s user think-time between calls costs one ~2-4s reconnect; the portfolio-sync connected-only lease will rarely find a live link anymore (@wabicai FYI — graceful degradation, but worth a look). A firmware issue for the underlying deaf-sleep behavior is being filed separately.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Desktop BLE acquire timeout still leaves a live stale session behind]

When the new 60s guard fires on the first BLE acquire, `device.mainId` is usually still unset because `Device.acquire()` only assigns it after `deviceConnector.acquire()` returns. That means this timeout cleanup skips `disconnect()`, even though the original transport call is still running.

Because `raceBleAcquire()` does not cancel that underlying `device.acquire()` promise, a late IPC reply can still come back afterward and let `Device.acquire()` set `deviceAcquired = true` and rebuild `DeviceCommands`. The retry path can then race or reuse the timed-out session instead of forcing the cold reconnect this hardening is trying to guarantee.

Please disconnect by the stable BLE descriptor id / connect id rather than `device.mainId`, and add an attempt or generation check so a timed-out acquire cannot mutate device state after the caller has already moved on.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1583,11 +1612,18 @@ async function disconnectDevice(deviceId: string): Promise<void> {
...
   });
+  if (peripheral.state === 'connected') {
+    logger?.warn('[NobleBLE] Physical disconnect did not confirm in time', { deviceId });
+  }
   cleanupDevice(deviceId, undefined, {
     cleanupConnection: true,
+    // macOS returns zero GATT services when a previously-disconnected
+    // peripheral object is reconnected; drop the discovery cache so the next
+    // connect resolves a fresh peripheral (direct connect by id, or scan).
+    cleanupDiscoveredCache: true,
```
originalix, line 1749: P2: [Clearing the discovered peripheral cache on every disconnect breaks the non-advertising fallback]

This new `cleanupDiscoveredCache: true` runs on every manual or idle disconnect, so the next reconnect no longer has the previous `Peripheral` object as a fallback. After that, `connectDevice()` can only recover through `tryDirectConnectById()` or `performTargetedScan()`.

That is not equivalent. `tryDirectConnectById()` returns `undefined` whenever `connectAsync` is unavailable or during its 15-second cooldown, and `performTargetedScan()` only succeeds if the device advertises again. A bonded device that stays silent between calls can now regress from reconnecting successfully to `DeviceNotFound`.

Please keep the last discovered peripheral until a replacement peripheral has been resolved successfully, or limit cache eviction to the backend cases where direct-connect is guaranteed to cover the reconnect path.

originalix (commented):
@originalix

Reviewed commit `881065c1f51a`.

- **P1 · Timed-out desktop BLE acquire can leave a stale session running**
- **P2 · Unconditional discovered-cache eviction breaks non-advertising reconnects**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#issuecomment-5330490106)

packages/core/src/core/index.ts
```diff
@@ -968,9 +1025,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+            device.acquire(method.payload.connectProtocol, {
+              forceProtocolDetection: method.payload.forceProtocolDetection,
+            }),
+            abortSignal
+          );
+        } catch (err) {
+          // A deadline hit means the transport is wedged mid-acquire; drop the
+          // link before the retry so it cold-connects instead of stacking a
+          // second connect onto the half-open one.
+          if (
```
originalix, line 1057: P1: [Timed-out desktop BLE acquire can leave a stale attempt active]

When the first desktop BLE acquire reaches this deadline, `Device.acquire()` is still awaiting the connector, so `device.mainId` has not been assigned and this cleanup guard skips `disconnect()`.

`raceBleAcquire()` rejects only its wrapper; the original acquire remains active while the retry loop starts another attempt. A late completion can still mark the Device acquired and replace `DeviceCommands`, allowing stale and current attempts to overlap.

Please make the Electron acquire attempt cancellable and invalidate its connection generation before starting a retry, using the stable BLE descriptor id for teardown rather than waiting for `mainId`.

packages/core/src/core/index.ts
originalix, line 424: P2: [The only-connect success path still leaks request tracing state]

When `onlyConnectBleDevice` succeeds, this removes the queue task but returns before `completeMethodRequestContext(method)` runs. The request therefore remains in the global and per-SDK active tracing maps until disposal, so repeated preconnections accumulate phantom running requests.

Please complete the method request context as part of this early success cleanup before returning.

originalix:
**Root cause found via a production control group — this changes the diagnosis (alpha.161).**

A 6.5.0 production log (old 1.1.31 transport) from the same Mac and the same Classic 1S was compared against this stack:

| | 6.5.0 (1.1.31) | this stack (1.2.x) |
|---|---|---|
| `Force reconnect` in one session | **32** | **0** (none in any log, ever) |
| `No services found` cycles | 16 | 0 |
| Protocol-deaf episodes | **0** | 74x err800 |
| Link ready -> device's first packet | **16/16 within ~1s**, including after a 6.8-min and a **15-min** idle | silent for 3s, then err800 loop |

So the device is NOT the problem, and neither is disconnect timing (my earlier hypotheses — unsubscribe-first, 20s hot disconnect — were both wrong; 20s still went deaf after 7m47s). The only behavioural difference that survives the comparison: 1.1.31 ran `forceReconnectPeripheral` **unconditionally on every cold setup**; #845 demoted it to a discovery-failure fallback, and since discovery on this device never fails, **the reset now never runs**. The link comes up, GATT resolves from cache, and the device answers nothing on a session it no longer serves — with no recovery path, because the old escape hatch (0 services -> deep cleanup -> fresh scan) is never triggered either.

This commit restores the old order in `setupConnectionAndDiscoverServices`: reset the link first, then discover, then fresh-scan fallback. A failed reset no longer aborts the attempt. Kept-alive links with an intact subscription return earlier in `connectDevice`, so reuse inside a workflow is untouched.

Cost, measured from 32 production samples: **+1.31s per cold setup** (disconnect ~5ms + reconnect 0.81s + the existing 500ms settle); cold connect goes ~1.6s -> ~2.9s. Kept-alive reuse: 0.

@wabicai — #845 demoted this deliberately as part of the Pro2 work; if there was a Pro2-specific reason the unconditional reset had to go, please flag it, since this restores it for all devices. A device-scoped variant is easy if needed. Field verification of the Classic 1S idle case is running now.

packages/core/src/core/index.ts
originalix, line 1058: P1: [Timed-out desktop BLE acquire can leave a stale attempt active]

When the first desktop BLE acquire reaches this timeout, `Device.acquire()` is still waiting on the Electron transport, so `device.mainId` has not been assigned yet and this cleanup guard skips `disconnect()`.

That means `raceBleAcquire()` rejects only its wrapper while the original acquire keeps running. The retry path then starts another acquire on the same Device generation, and a late completion of the first attempt can still mark the Device acquired and replace `DeviceCommands` underneath the retry.

Please make the desktop acquire attempt cancellable, or invalidate this Device generation and tear it down by the stable BLE descriptor id before starting any retry.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1305,26 +1334,28 @@ async function setupConnectionAndDiscoverServices(
...
+  // link with an intact subscription returns before reaching here, so reuse
+  // inside a workflow is unaffected.
   try {
-    return await discoverServicesAndCharacteristicsWithRetry(peripheral, deviceId);
-  } catch (initialError) {
-    logger?.info('[NobleBLE] Direct service discovery failed, retrying after reconnect', {
-      deviceId,
-      error: initialError,
-    });
+    await forceReconnectPeripheral(peripheral, deviceId);
```
originalix, line 1412: P2: [Forced reconnect still reconnects before the old link is confirmed down]

This path now sends every cold setup through `forceReconnectPeripheral()`, but that helper still treats disconnect as complete after the old 250ms timeout.

The same file now documents that 250ms is too short to prove physical teardown on macOS. If the callback does not arrive in time, the reconnect starts while CoreBluetooth is still closing the old link, so the new reset path can recreate the same stale-GATT state it is meant to clear.

Please wait for a confirmed disconnect before reconnecting here, or fall back to a fresh scan when the disconnect does not settle.

packages/core/src/core/index.ts
originalix, line 424: P2: [The only-connect success path still leaks request tracing state]

When `onlyConnectBleDevice` succeeds, this branch now releases the queue task but still returns before `completeMethodRequestContext(method)` runs.

That leaves the request registered in the global and per-SDK active tracing maps until disposal, so repeated preconnect calls accumulate phantom active work and make later queue diagnostics misleading.

Please complete the request context before returning from this early-success path, ideally by reusing the normal successful-call finalization.

packages/core/src/core/index.ts
originalix, line 1058: P1: [Timed-out desktop BLE acquire can leave a stale attempt active]

When the first desktop acquire reaches this deadline, `Device.acquire()` can still be waiting for the Electron transport, so `device.mainId` has not been assigned and this cleanup guard is skipped.

`raceBleAcquire()` then rejects only its wrapper and the retry starts while the original attempt remains active. If that attempt completes later, it can still assign `mainId`, mark the Device acquired, and replace `DeviceCommands` during the newer attempt.

Please move this deadline into the Electron transport acquire so it aborts the underlying IPC/native connection, invalidates the Device attempt, and completes teardown by the stable connect ID before Core retries.

packages/core/src/core/index.ts
originalix, line 424: P2: [Only-connect success still leaks request tracing state]

When `onlyConnectBleDevice` succeeds, this branch releases the queue task but returns before `completeMethodRequestContext(method)` runs.

The request therefore remains in the global and per-SDK active maps until disposal, so repeated preconnect calls accumulate phantom active work and make later diagnostics misleading.

Please complete the method request context before releasing the task and returning from this branch.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Timed-out acquire can mutate a newer connection attempt]

When the deadline or polling timeout wins, `raceBleAcquire` rejects but the underlying `Device.acquire` keeps running. The cleanup here only disconnects and marks the transport disconnected; it does not invalidate the acquire generation. If the old acquire resolves after the retry starts, `Device.acquire` can still commit `mainId`, `deviceAcquired`, commands, and protocol state for the stale link, corrupting the newer session.

Invalidate the abandoned acquire with the same generation/interruption mechanism used for cancellation, and guard the state-commit section before retrying so a late completion cannot overwrite the active connection.

originalix, line 1047: P1: [Timed-out acquire can mutate a newer connection attempt]

When the deadline or abort wins, `raceBleAcquire` rejects but the underlying `Device.acquire` keeps running. The timeout cleanup only disconnects the current link and marks transport state; it does not invalidate the acquire generation. If the old acquire resolves after the retry starts, `Device.acquire` can still commit `mainId`, `deviceAcquired`, commands, and protocol state for the stale link, corrupting the newer session.

Invalidate the abandoned acquire with the same interruption or generation mechanism used for cancellation, and guard the state-commit section so a late completion cannot overwrite the active connection.

originalix: Fixed in `3dd1b964d8` (published as 1.2.0-alpha.166).

`connectDeviceForBle` now calls `device.beginConnectionAttempt()` in the deadline/abort catch, before the disconnect attempt. That supersedes the abandoned acquire, so when it settles later the generation guard already in `Device.acquire` (`interruptedAttempt === attempt || connectionAttempt !== attempt`) drops the link it built and throws `DeviceInterruptedFromUser` instead of committing `mainId`, `deviceAcquired`, `commands` or protocol state over the newer attempt. The bump also covers the case you called out where `mainId` is still unset, so the existing disconnect guard is skipped.

Your stronger suggestion — moving the deadline into the Electron transport so it aborts the underlying IPC — is the better end state and I have not done it here; this keeps the fix inside Core where the retry lives. Happy to follow up with the transport-level abort in a separate change.

Duplicate threads for this same finding are being resolved together; this one carries the resolution.

packages/core/src/core/index.ts
originalix: Fixed in `3dd1b964d8` (1.2.0-alpha.166): the `onlyConnectBleDevice` branch now calls `completeMethodRequestContext(method)` before `requestQueue.releaseTask()` and returning, so the request leaves the global and per-SDK active maps instead of lingering until disposal. Repeated preconnect calls no longer accumulate phantom active work in diagnostics.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Fixed in `3dd1b964d8` (1.2.0-alpha.166), close to your first suggestion. `cleanupDevice` no longer drops the evicted `Peripheral` outright: it moves it into a process-lifetime `stalePeripherals` map, and `connectDevice` uses it only as the last resort, after both a targeted scan and `tryDirectConnectById` come up empty (and clears it as soon as a fresh peripheral is resolved). A bonded device that stays silent — or a connect-by-id sitting in its 15s cooldown — therefore keeps the old reconnect path instead of regressing to `DeviceNotFound`, while a stale object is never *preferred* again, which is what the eviction was for: reconnecting a retrieved/cached peripheral is what produces the link where GATT resolves from cache but the device answers no protocol traffic.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Fixed in `3dd1b964d8` (1.2.0-alpha.166): `forceReconnectPeripheral` now waits on `BLE_DISCONNECT_CONFIRM_TIMEOUT_MS` (3s) instead of the 250ms cleanup cap, and logs a warning when the peripheral is still `connected` after that window so an unconfirmed teardown stops being invisible. Field logs show this disconnect completing in 1–5ms, so the longer cap costs nothing on the normal path and only buys time on the slow one.

packages/core/src/core/index.ts
originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

originalix: Duplicate of the acquire-timeout finding that successive review rounds re-filed on this same code (10 threads for one issue). Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — the resolution is written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the `onlyConnectBleDevice` bookkeeping finding re-filed across review rounds. Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — resolution written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the `onlyConnectBleDevice` bookkeeping finding re-filed across review rounds. Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — resolution written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Duplicate of the `onlyConnectBleDevice` bookkeeping finding re-filed across review rounds. Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — resolution written up on the newest thread. Resolving this one to cut the noise.

originalix: Duplicate of the `onlyConnectBleDevice` bookkeeping finding re-filed across review rounds. Fixed in `3dd1b964d8` / 1.2.0-alpha.166 — resolution written up on the newest thread. Resolving this one to cut the noise.

packages/core/src/core/index.ts
originalix: Correction: this fix is **deferred, not shipped**. It landed in `d6be62d0f` (published as 1.2.0-alpha.166) and has now been reverted in `72e137fca`.

Reason: this file's connection path is currently under field validation against a 6.5.0 control build — the cold-connect route, the forced reset and the cache eviction were each derived from device logs, and changing them mid-validation would invalidate the comparison. The finding itself stands and is queued to land once the root cause is confirmed.

Nothing under test was affected: the app PR pins 1.2.0-alpha.165, which never contained these changes.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Correction: this fix is **deferred, not shipped**. It landed in `d6be62d0f` (published as 1.2.0-alpha.166) and has now been reverted in `72e137fca`.

Reason: this file's connection path is currently under field validation against a 6.5.0 control build — the cold-connect route, the forced reset and the cache eviction were each derived from device logs, and changing them mid-validation would invalidate the comparison. The finding itself stands and is queued to land once the root cause is confirmed.

Nothing under test was affected: the app PR pins 1.2.0-alpha.165, which never contained these changes.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Correction: this fix is **deferred, not shipped**. It landed in `d6be62d0f` (published as 1.2.0-alpha.166) and has now been reverted in `72e137fca`.

Reason: this file's connection path is currently under field validation against a 6.5.0 control build — the cold-connect route, the forced reset and the cache eviction were each derived from device logs, and changing them mid-validation would invalidate the comparison. The finding itself stands and is queued to land once the root cause is confirmed.

Nothing under test was affected: the app PR pins 1.2.0-alpha.165, which never contained these changes.

packages/core/src/core/index.ts
```diff
@@ -968,9 +1028,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+      if (useAcquireGuards && abortSignal?.aborted) {
+        throw ERRORS.TypedError(HardwareErrorCode.CallQueueActionCancelled);
+      }
+      if (!useAcquireGuards) {
+        await device.acquire(method.payload.connectProtocol, {
+          forceProtocolDetection: method.payload.forceProtocolDetection,
+        });
+      } else {
+        try {
+          await raceBleAcquire(
```
originalix, line 1047: P1: [Timed-out acquire can mutate a newer connection attempt]

When the deadline or abort wins, `raceBleAcquire` rejects but the underlying `Device.acquire` keeps running. The timeout cleanup only disconnects the current link when `device.mainId` is already set; on a first acquire there may be no session to disconnect, and the acquire generation is not invalidated. If the old acquire resolves after the retry starts, `Device.acquire` can still commit `mainId`, `deviceAcquired`, commands, and protocol state for the stale link, corrupting the newer session.

Invalidate the abandoned acquire with the same interruption or generation mechanism used for cancellation, and guard the state-commit section so a late completion cannot overwrite the active connection.

packages/core/src/core/index.ts
```diff
@@ -968,9 +1028,43 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
...
+            device.acquire(method.payload.connectProtocol, {
+              forceProtocolDetection: method.payload.forceProtocolDetection,
+            }),
+            abortSignal
+          );
+        } catch (err) {
+          // A deadline hit means the transport is wedged mid-acquire; drop the
+          // link before the retry so it cold-connects instead of stacking a
+          // second connect onto the half-open one.
+          if (
```
originalix, line 1057: P1: [Acquire timeout retries while the original BLE acquire is still in flight]

When the 60-second race expires, only the wrapper promise is rejected; the underlying `device.acquire()` continues running. `Device.mainId` is assigned after that promise resolves, so this guard normally has no session to disconnect and the recursive retry starts a second acquire on the same device. A late response from the first acquire can then recreate commands or mark the device acquired after the retry has moved on.

Please invalidate/cancel the underlying acquire (or advance the connection-attempt generation so late completion is discarded) and wait for that operation to be safely quarantined before starting a retry.

packages/core/src/core/index.ts
```diff
@@ -415,6 +415,13 @@ const onCallDevice = async (
   if (method.payload?.onlyConnectBleDevice) {
```
originalix, line 415: P2: [Connect-only success path still skips method disposal]

This early return now completes tracing and removes the request task, but it still bypasses the common cleanup block that calls `method.dispose()`. Repeated successful connect-only/preconnect calls can therefore retain method-owned resources or listeners.

Dispose the method before returning from this branch, or route the response through the same cleanup/finally path used by normal calls.

packages/core/src/core/index.ts
originalix, line 1057: P1: [Acquire timeout retries while the original BLE acquire is still in flight]

When the 60-second race expires, only the wrapper promise is rejected; the underlying `device.acquire()` continues running. `Device.mainId` is assigned after that promise resolves, so this guard normally has no session to disconnect and the recursive retry starts a second acquire on the same device. A late response from the first acquire can then recreate commands or mark the device acquired after the retry has moved on.

Please invalidate or cancel the underlying acquire, or advance the connection-attempt generation so late completion is discarded, and quarantine that operation before starting a retry.

packages/core/src/core/index.ts
originalix, line 415: P2: [Connect-only success path still skips method disposal]

This early return now completes tracing and removes the request task, but it still bypasses the common cleanup block that calls `method.dispose()`. Repeated successful connect-only/preconnect calls can therefore retain method-owned resources or listeners.

Dispose the method before returning from this branch, or route the response through the same cleanup/finally path used by normal calls.

packages/core/src/core/index.ts
originalix, line 1057: P1: [Acquire timeout retries while the original BLE acquire is still in flight]

When the deadline or cancellation fires, only the wrapper promise is rejected; the underlying `device.acquire()` continues running. `Device.mainId` is usually assigned after that promise resolves, so this cleanup often has no session to disconnect and the retry starts a second acquire on the same device. A late response from the first acquire can then recreate commands or mark the device acquired after the retry has moved on.

Please cancel or invalidate the underlying acquire and quarantine late completion with the connection-attempt generation before starting a retry.

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -487,13 +487,20 @@ export default class ElectronBleTransport {
...
+    // already does on iOS: the caller reads the protocol off its own device
+    // record, so probing re-asks a question that is already answered and adds a
+    // round trip to every cold connect. It also fails in a way that costs the
+    // session: a device whose protocol session has stalled ignores the probe
+    // frame, and the timeout became a protocol-mismatch error that stopped Core
+    // from ever sending Initialize — the one frame such a device still answers,
+    // and how it gets revived. A declared V2 keeps probing, matching iOS, so a
+    // USB-priority "link disabled" surfaces here rather than as an unmapped
+    // error later. An undeclared protocol still goes through full detection.
     if (expectedProtocol === 'V1') {
```
originalix, line 500: P1: [Protocol V1 is accepted without active device confirmation]

This branch sets V1 solely from `expectedProtocol` and skips any device response. After a reconnect, firmware transition, or device replacement, that metadata can be stale; the transport will label a V2 endpoint as V1 and send the wrong framing, causing communication or firmware-reconnect failures.

Please retain the V1 probe unless V1 was confirmed on the current connection generation, using metadata only to choose probe order.

packages/core/src/core/index.ts
originalix, line 415: P2: [Connect-only success bypasses method disposal]

This early return now completes tracing and removes the request task, but it still bypasses the common cleanup path that calls `method.dispose()`. Repeated successful connect-only/preconnect calls can retain method-owned resources or listeners.

Please dispose the method before returning from this branch, or route the response through the same cleanup/finally path used by normal calls.

originalix (commented):
@originalix

Reviewed commit `294d4c3ce64a`.

- **P1 · BLE notification diagnostics expose protocol payload bytes**
- **P1 · Protocol V1 is accepted without active device confirmation**
- **P1 · Acquire timeout retries while the original BLE acquire is still active**
- **P2 · Cold reconnect can begin before physical disconnect completes**
- **P2 · Connect-only success bypasses method disposal**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#issuecomment-5330490106)

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -703,6 +733,17 @@ async function transmitHexDataToDevice(
   );
   const pacingDelayMs = resolveNobleBleWritePacingDelay(options);
 
+  // Diagnostics: the frame header alone (report id + magic + message type +
+  // length) identifies which protocol frame went out, without logging payload.
+  logger?.info('[NobleBLE] Write frame', {
+    deviceId,
+    bytes: toBuffer.length,
+    head: toBuffer.subarray(0, 8).toString('hex'),
```
originalix: P1: [BLE frame diagnostics expose protocol payload bytes]

When a Protocol V1 request spans multiple packets, this hook is called separately for each packet. Only the first packet has the message header; later packets contain raw protobuf payload, so `head` can persist signing or wallet request bytes in desktop logs.

Please remove the hexadecimal preview and retain only non-sensitive scalars such as byte count, packet capacity, MTU, and pacing.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1732,6 +1880,11 @@ async function subscribeNotifications(
     });
 
     notifyCharacteristic.on('data', (data: Buffer) => {
+      logger?.debug('[NobleBLE] Notify frame', {
+        deviceId,
+        bytes: data.length,
+        head: data.subarray(0, 8).toString('hex'),
```
originalix: P1: [BLE frame diagnostics expose protocol payload bytes]

When a Protocol V1 response spans multiple notifications, this listener also receives continuation packets containing raw protobuf payload rather than a frame header. Logging their first eight bytes can persist address, wallet, signing, or device response data.

Please remove the byte preview and log only non-sensitive values such as packet length and lifecycle state.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix, line 500: P1: [Protocol V1 is accepted without active device confirmation]

On reconnect, `expectedProtocol` can come from persisted or earlier descriptor state. This branch converts that metadata directly into detected V1 without receiving a valid V1 response; if the endpoint or device state has changed, subsequent calls are routed through unconfirmed V1 framing and communication fails.

Please retain the active V1 probe and use `expectedProtocol` only as the strict result to validate.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -2120,31 +2120,23 @@ export default class ReactNativeBleTransport {
...
-      this.deviceProtocol.set(uuid, expectedProtocol);
+    // A declared V1 is taken at face value on every platform, as iOS has done
+    // since protocol probing arrived: the caller reads the protocol off its own
+    // device record, so the probe re-asks a question that is already answered
+    // and costs a round trip on every acquire. Expected V2 must still Ping so
+    // USB-priority `link disabled` surfaces here instead of as a later unmapped
+    // RuntimeError. sessionProtocols is stamped too — the iOS path skipped it,
+    // which left the firmware-install reconnect (skipProtocolProbe) without the
+    // confirmed protocol it requires.
+    if (expectedProtocol === 'V1') {
```
originalix, line 2133: P1: [Protocol V1 is accepted without active device confirmation]

When Android or iOS reconnects with stale `expectedProtocol` metadata, this branch stamps both protocol maps without receiving a V1 response. The adapter can then route calls through V1 framing even though the active endpoint was never confirmed.

Please restore active V1 confirmation and populate `deviceProtocol` and `sessionProtocols` only after that probe succeeds.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Acquire deadline allows overlapping native acquires]

If the deadline wins, only `raceBleAcquire` rejects; the underlying `device.acquire()` keeps running. `device.mainId` is normally assigned only after that promise resolves, so the timeout cleanup often cannot disconnect it, and the recursive retry starts another acquire while the expired operation can still complete and mutate Device state.

Please invalidate the expired connection-attempt generation and disconnect by the known BLE connect ID before starting another acquire.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1305,26 +1410,42 @@ async function setupConnectionAndDiscoverServices(
...
+  // link with an intact subscription returns before reaching here, so reuse
+  // inside a workflow is unaffected.
   try {
-    return await discoverServicesAndCharacteristicsWithRetry(peripheral, deviceId);
-  } catch (initialError) {
-    logger?.info('[NobleBLE] Direct service discovery failed, retrying after reconnect', {
-      deviceId,
-      error: initialError,
-    });
+    await forceReconnectPeripheral(peripheral, deviceId);
```
originalix, line 1412: P2: [Cold reconnect can start before physical disconnect completes]

Every cold setup now enters `forceReconnectPeripheral`, but that helper resolves its disconnect after the 250 ms cleanup timeout and immediately calls `connect()` on the same object. When Noble or CoreBluetooth is still disconnecting, the reconnect can reuse the stale GATT session this reset is intended to clear.

Please wait for confirmed physical disconnection using the longer teardown deadline and resolve a fresh peripheral if teardown cannot be confirmed.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: P1: [BLE frame diagnostics expose protocol payload bytes]

When a request spans multiple BLE packets, this hook is called separately for each packet. Continuation packets begin with raw protobuf payload rather than a frame header, so `head` can persist signing or wallet request bytes in desktop logs. The notification listener at the same time has the same issue for response continuations.

Please remove the hexadecimal previews and retain only non-sensitive scalars such as byte count, packet capacity, MTU, and pacing.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix, line 500: P1: [Protocol V1 is accepted without active device confirmation]

On reconnect, `expectedProtocol` can come from persisted or earlier descriptor state. This branch converts that metadata directly into detected V1 without receiving a valid V1 response; if the endpoint or device state has changed, subsequent calls are routed through unconfirmed V1 framing and communication fails.

Please retain the active V1 probe and use `expectedProtocol` only as the strict result to validate.

packages/hd-transport-react-native/src/index.ts
originalix, line 2133: P1: [Protocol V1 is accepted without active device confirmation]

When Android or iOS reconnects with stale `expectedProtocol` metadata, this branch stamps both protocol maps without receiving a V1 response. The adapter can then route calls through V1 framing even though the active endpoint was never confirmed.

Please restore active V1 confirmation and populate `deviceProtocol` and `sessionProtocols` only after that probe succeeds.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Acquire deadline allows overlapping native acquires]

If the deadline wins, only `raceBleAcquire` rejects; the underlying `device.acquire()` keeps running. `device.mainId` is normally assigned only after that promise resolves, so the timeout cleanup often cannot disconnect it, and the recursive retry starts another acquire while the expired operation can still complete and mutate Device state.

Please invalidate the expired connection-attempt generation and disconnect by the known BLE connect ID before starting another acquire.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Declining, with the reasoning on record. `discoveredDevices` is the handler's own Map — noble keeps its own peripheral registry (`noble._peripherals`) that nothing in this file ever clears, so deleting our entry does not destroy the Peripheral object, it only changes which code path the next connect takes. Connect-by-id and a cached `peripheral.connect()` converge on the same binding call (`bindings.connect(uuid)`), and none of the three bindings require a live advertisement (macOS falls back to `retrievePeripheralsWithIdentifiers`, Windows to `FromBluetoothAddressAsync`, hci-socket to `createLeConn`). The non-advertising case is handled by the cold-connect ladder: the scan route comes up empty and falls through to `connectById()`. The cache was also never durable — `enumerateDevices` clears it on every device-list refresh and the poweredOff handler clears it too. Keeping `cleanupDiscoveredCache: true` is what removed the field failure this PR exists for: reusing a stale peripheral after an idle teardown made `discoverServices` return 0 services on Classic 1S, hardware-verified across several TestFlight builds. What would change our mind: a field log where connect-by-id returns DeviceNotFound while a retained `peripheral.connect()` on the same id succeeds.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Valid observation, deliberately deferred. Two confirm windows do coexist today: `disconnectDevice` waits on `BLE_DISCONNECT_CONFIRM_TIMEOUT_MS` (3000ms) with a logged miss, while `forceReconnectPeripheral` still resolves its teardown after `BLE_CLEANUP_TIMEOUT` (250ms) and reconnects immediately. Recent field logs (build 2026082180) add a sharper signal: on Classic 1S, service discovery after a force-reconnect returned 0 services in 3/3 cold connects, while discovery on a freshly scanned peripheral succeeded 4/4 — so this helper is under active investigation as a possible cause of the failure it was meant to cure, not just an insufficient confirm window. That question needs a controlled A/B on hardware (skip force-reconnect on the first attempt and observe first-discovery results) before we change the sequence: this reconnect flow was validated frame-by-frame on devices across many builds, and we do not modify validated communication ordering on inference. Tracked together with the alpha.181 field-observation follow-ups.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827110054 . Resolving as duplicate.

packages/core/src/core/index.ts
originalix: Acknowledged as a real (if now rare) race: when the deadline or abort wins, the underlying `device.acquire()` keeps running and a late settlement can still commit state onto a newer attempt, since the timeout cleanup is guarded on `device.mainId` which is normally assigned only after the acquire resolves. Deferring the fix deliberately: 1.2.0-alpha.181 fixed the main-process defect that made the 60s deadline fire at all (an IPC promise that could never settle), so the deadline is now a rarely-armed backstop, and the candidate fix (a dedicated acquire epoch checked before committing `mainId`/commands state) touches the acquire path we have just finished stabilizing through several hardware-validated TestFlight rounds. We will land the epoch guard in a follow-up once the current transport build has accumulated field time, rather than piggybacking it onto this PR.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

packages/core/src/core/index.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

packages/core/src/core/index.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

packages/core/src/core/index.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Resolving as duplicate.

packages/core/src/core/index.ts
originalix: The control-flow observation is accurate — the early return at core/index.ts:415 does bypass the shared cleanup block that calls `method.dispose()`. But `BaseMethod.dispose()` is an empty method (`dispose() {}`, BaseMethod.ts:388) and no BLE-path method subclass overrides it with resource cleanup, so nothing is retained by skipping it. The early return already performs the bookkeeping that has observable effects (`completeMethodRequestContext`, `requestQueue.releaseTask`). Declining the change as there is no leak to fix; if a future method subclass gains a real `dispose`, this branch should be revisited then.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3823668463 . Resolving as duplicate.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3823668463 . Resolving as duplicate.

packages/hd-transport-react-native/src/index.ts
originalix: Declining the probe restoration; the declared-V1 fast path stays. We searched specifically for a reachable case where a declared V1 meets a V2 endpoint: bootloader mode keeps its stored V2 protocol (Device.ts strictProtocol resolution), BLE connectIds are per-device in the app DB so a device swap gets a fresh record, and there is no V1→V2 firmware transition for shipped V1 devices. The declaration also remains checked: `DeviceConnector.acquire` still throws `Device protocol mismatch` when a probe result contradicts it, and `forceProtocolDetection` still nulls the declaration and runs a full probe. The probe this branch skips was actively harmful in the field, not just redundant: a Classic 1S in the post-idle-disconnect state ignores the V1 GetFeatures probe entirely (3s timeout, err 800) while still answering Initialize, so probing blocked recovery — that behavior is hardware-verified and is the reason iOS carved this out first (dedd52aaa). Related hardening that did come out of this review: 1.2.0-alpha.182 (commit 2e4b24536) removes the `sessionProtocols` stamp this branch briefly added on react-native, so forced detection is never narrowed by an unverified declaration — the map again only records protocols the device actually answered on.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Resolving as duplicate.

packages/hd-transport-react-native/src/index.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Resolving as duplicate.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Resolving as duplicate.

originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Resolving as duplicate.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Confirmed and fixed in 1.2.0-alpha.182 (commit aaf7b7ed2). Both hooks fire once per BLE packet, not per frame, and continuation packets start in raw protobuf payload — field logs showed a device serial (`CLA45F00…`) and a firmware version string landing in the desktop log through the notify preview. The `head` field now goes through `frameHeadPreview()`, which prints bytes only for a real frame header: Protocol V1 via the existing `isHeaderChunk` check from `@onekeyfe/hd-shared` (8 bytes: report id + magic + type + length), Protocol V2 via the `0x5A` SOF (7 bytes — byte 7 is already payload). Continuation packets log scalars only (`bytes`, `packetCapacity`, `mtu`, `pacingDelayMs`). Diagnostics-only change; the transmit path and packet contents are untouched.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978135 . Resolving as duplicate.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Duplicate of this finding from an earlier review pass; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978135 . Resolving as duplicate.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix, line 500: P1: [Protocol V1 is accepted without active device confirmation]

When `expectedProtocol` comes from stale persisted or previous descriptor state, this branch returns V1 without receiving a response from the active device. Core then stores that result as the detected protocol, so subsequent traffic is routed through V1 framing and the mismatch appears only as a later timeout or command failure.

Please keep `expectedProtocol` as the strict result to validate and update `deviceProtocol` only after a non-destructive V1 probe receives a valid response.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -2120,31 +2120,24 @@ export default class ReactNativeBleTransport {
...
+    // since protocol probing arrived: the caller reads the protocol off its own
+    // device record, so the probe re-asks a question that is already answered
+    // and costs a round trip on every acquire. Expected V2 must still Ping so
+    // USB-priority `link disabled` surfaces here instead of as a later unmapped
+    // RuntimeError. sessionProtocols is deliberately NOT stamped here: that map
+    // records protocols the device actually answered on (it gates the
+    // trustSessionProtocol narrowing in forced detection), and this branch has
+    // received no response. skipProtocolProbe does not need it either — its only
+    // caller is the V2 firmware-install reconnect.
+    if (expectedProtocol === 'V1') {
```
originalix, line 2133: P1: [Protocol V1 is accepted without active device confirmation]

If a mobile reconnect supplies stale `expectedProtocol` metadata, this branch publishes V1 without receiving any V1 response from the active endpoint. The returned protocol becomes Core's detected protocol and later calls are routed through unconfirmed V1 framing, producing delayed communication failures instead of a strict mismatch here.

Please restore the V1 response probe and set `deviceProtocol` only after that probe validates the expectation.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Acquire deadline allows overlapping native acquires]

When this deadline wins, only `raceBleAcquire` rejects; the underlying `device.acquire()` keeps running. Because `mainId` is normally assigned after that promise resolves, the cleanup guard usually skips the first in-flight link, and the recursive retry starts another acquire while the expired one can still complete and mutate Device commands or protocol state.

Please assign each acquire a cancellable generation, invalidate that generation on deadline, disconnect using the known BLE connect ID, and await teardown before retrying so a late settlement cannot update Device state.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1305,26 +1434,45 @@ async function setupConnectionAndDiscoverServices(
...
```
originalix, line 1412: P2: [Cold reconnect can start before physical disconnect completes]

Every cold setup now enters `forceReconnectPeripheral`, but that helper resolves its disconnect wait after 250 ms and immediately reconnects the same peripheral. When Noble or CoreBluetooth is still tearing down, the new connection can reuse stale GATT state or be interrupted by the delayed disconnect, defeating this recovery path.

Please use the confirmed disconnect deadline, verify that the peripheral is no longer connected, and obtain a fresh peripheral before reconnecting if teardown does not confirm.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix, line 500: P1: [Protocol V1 is accepted without active device confirmation]

When `expectedProtocol` comes from stale persisted or previous descriptor state, this branch returns V1 without receiving a response from the active device. Core then stores that result as the detected protocol, so subsequent traffic is routed through V1 framing and the mismatch appears only as a later timeout or command failure.

Please restore the active V1 probe and update `deviceProtocol` only after that probe validates the expectation.

packages/hd-transport-react-native/src/index.ts
originalix, line 2133: P1: [Protocol V1 is accepted without active device confirmation]

If a mobile reconnect supplies stale `expectedProtocol` metadata, this branch publishes V1 without receiving any V1 response from the active endpoint. The returned protocol becomes Core's detected protocol and later calls are routed through unconfirmed V1 framing, producing delayed communication failures instead of a strict mismatch here.

Please restore the active V1 probe and set `deviceProtocol` only after that probe validates the expectation.

packages/core/src/core/index.ts
originalix, line 1047: P1: [Acquire deadline allows overlapping native acquires]

When this deadline wins, only `raceBleAcquire` rejects; the underlying `device.acquire()` keeps running. Because `mainId` is normally assigned after that promise resolves, cleanup can miss the first in-flight link, and the recursive retry starts another acquire while the expired one can still complete and mutate Device commands or protocol state.

Please invalidate the Device connection-attempt generation, hard-disconnect `method.connectId`, and await teardown before retrying so a late settlement cannot update Device state.

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -1305,26 +1399,45 @@ async function setupConnectionAndDiscoverServices(
...
```
originalix, line 1412: P2: [Cold reconnect can start before physical disconnect completes]

Every cold setup now enters `forceReconnectPeripheral`, but that helper resolves its disconnect wait after 250 ms and immediately reconnects the same peripheral. When Noble or CoreBluetooth is still tearing down, the new connection can reuse stale GATT state or be interrupted by the delayed disconnect, defeating this recovery path.

Please make the helper wait for the confirmed disconnect deadline and resolve a fresh peripheral instead of reconnecting this object when teardown does not confirm.

packages/hd-transport-react-native/src/index.ts
originalix: Re-raised on a push that did not touch this code path — the adjudication stands: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Since then, 1.2.0-alpha.182 (commit 2e4b24536) additionally removed the `sessionProtocols` stamp from this branch, so an unverified declaration can no longer narrow forced detection; the map again records only protocols the device actually answered on. The declared-V1 fast path itself is hardware-validated and stays. Resolving as a duplicate of the consolidated thread.

packages/core/src/core/index.ts
originalix: Re-raised duplicate; consolidated response and deferral rationale: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Status unchanged: acknowledged as a real but narrow race, deliberately deferred by maintainer decision — alpha.181 removed the only known trigger of the 60s deadline (it fired 0 times in the latest field session), and the epoch-guard fix touches the acquire path we just finished stabilizing on hardware. Reopening condition on record: field evidence of a late acquire committing state onto a newer attempt. Resolving as duplicate.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Re-raised duplicate; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827110054 . Status unchanged and sharpened: field data (build 2026082180) shows service discovery after a force-reconnect returned 0 services 3/3 on Classic 1S while a freshly scanned peripheral succeeded 4/4, so this helper is under investigation as a possible cause of the failure it was meant to cure. The fix will be decided by a controlled hardware A/B (possibly skipping force-reconnect entirely for this family), not by aligning the confirm window on inference — validated communication ordering is not changed without device evidence. Resolving as duplicate.

packages/hd-transport-web-device/src/electron-ble-transport.ts
originalix: Re-raised on a push that did not touch this code path — the adjudication stands: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Since then, 1.2.0-alpha.182 (commit 2e4b24536) additionally removed the `sessionProtocols` stamp from this branch, so an unverified declaration can no longer narrow forced detection; the map again records only protocols the device actually answered on. The declared-V1 fast path itself is hardware-validated and stays. Resolving as a duplicate of the consolidated thread.

packages/hd-transport-react-native/src/index.ts
originalix: Re-raised on a push that did not touch this code path — the adjudication stands: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827978943 . Since then, 1.2.0-alpha.182 (commit 2e4b24536) additionally removed the `sessionProtocols` stamp from this branch, so an unverified declaration can no longer narrow forced detection; the map again records only protocols the device actually answered on. The declared-V1 fast path itself is hardware-validated and stays. Resolving as a duplicate of the consolidated thread.

packages/core/src/core/index.ts
originalix: Re-raised duplicate; consolidated response and deferral rationale: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827979339 . Status unchanged: acknowledged as a real but narrow race, deliberately deferred by maintainer decision — alpha.181 removed the only known trigger of the 60s deadline (it fired 0 times in the latest field session), and the epoch-guard fix touches the acquire path we just finished stabilizing on hardware. Reopening condition on record: field evidence of a late acquire committing state onto a newer attempt. Resolving as duplicate.

packages/hd-transport-electron/src/noble-ble-handler.ts
originalix: Re-raised duplicate; consolidated response: https://github.com/OneKeyHQ/hardware-js-sdk/pull/888#discussion_r3827110054 . Status unchanged and sharpened: field data (build 2026082180) shows service discovery after a force-reconnect returned 0 services 3/3 on Classic 1S while a freshly scanned peripheral succeeded 4/4, so this helper is under investigation as a possible cause of the failure it was meant to cure. The fix will be decided by a controlled hardware A/B (possibly skipping force-reconnect entirely for this family), not by aligning the confirm window on inference — validated communication ordering is not changed without device evidence. Resolving as duplicate.

zhaono1 (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

wabicai (approved):

packages/core/src/core/index.ts
```diff
@@ -1010,7 +1104,7 @@ async function connectDeviceForBle(method: BaseMethod, device: Device, retryCoun
       const nextRetry = retryCount + 1;
       Log.debug(`Bluetooth connection will retry, retry count: ${nextRetry}`);
       await wait(3000);
-      await connectDeviceForBle(method, device, nextRetry);
+      await connectDeviceForBle(method, device, abortSignal, nextRetry);
```
sidmorizon, line 1107: # 🟠 P1 acquire 的 60s 期限被重试放大到约 438s

## 问题

`raceBleAcquire` 超时后抛出的是 `ERRORS.TypedError(HardwareErrorCode.BleTimeoutError, ...)`，而 `isRetryableBleConnectionError()` 对 `BleTimeoutError` 直接返回 `true`（`index.ts:943`）。于是本行的递归重试被触发：`retryCount < 6` 允许 `retryCount = 0..6` 共 **7 次** acquire，每次之间 `await wait(3000)`。

外层也没有兜底预算：`ensureConnected` 在 `await connectDeviceForBle(...)` 之前已经 `clearTimeout(timer)`（`index.ts:1200-1202`），所以 `TIME_OUT`（默认 `10000`）那层轮询超时对这段完全不生效。

## 影响

在本 PR 要修的那个失败模式下（Electron 主进程 IPC 回复永不到达、`device.acquire()` 永不 settle），每次重试都会重新等满 `BLE_ACQUIRE_DEADLINE_MS`：最坏 `7 × 60s + 6 × 3s ≈ 438s`。这比 PR 描述中作为修复目标的 `306s` 现场卡死更长，"60s 期限"对调用方并不成立；同时这 7 次尝试各自留下一个仍在运行的 `device.acquire()`。

## 建议

让期限错误对重试循环终结，或给整段 acquire 一个累计预算。例如给期限错误加独立标记并在重试判定中排除：

```ts
// raceBleAcquire 内
const timeoutError = ERRORS.TypedError(
  HardwareErrorCode.BleTimeoutError,
  `BLE acquire exceeded ${BLE_ACQUIRE_DEADLINE_MS}ms deadline`
);
(timeoutError as { bleAcquireDeadline?: boolean }).bleAcquireDeadline = true;

// connectDeviceForBle catch 内
const hitAcquireDeadline = (err as { bleAcquireDeadline?: boolean }).bleAcquireDeadline === true;
if (!hitAcquireDeadline && isRetryableBleConnectionError(method, err) && retryCount < 6) { ... }
```

或在 `connectDeviceForBle` 首次进入时记录 `startedAt`，重试前校验 `Date.now() - startedAt` 未超过总预算。

[Claude session ↗](https://claude.ai/code/session_0153pNwiNRrAHfrvhF1qHzjb)

---

PR #890, opened 2026-08-19
fix: canonicalize Pro2 BLE names and simplify Protocol V2 firmware updates

wabicai:
## Summary

SDK fixes for OneKey Pro 2 BLE name collisions and Protocol V2 firmware/resource updates.

## BLE names

- Canonicalize and compare Pro2 BLE names as `Pro 2 XXXX`.
- Still accept legacy compact `Pro2 XXXX`.
- Do not rewrite OneKey Pro names that only happen to start with `Pro 2` (for example `Pro 22D8`).
- Apply the same comparison in search, Protocol V2 device info/features, state mapping, device-name helpers, and CLI `--device-name`.

## FirmwareUpdateV4

There are two update paths:

1. **Remote (app-monorepo):** `firmwareUpdateV4({ preparedPlan, hostBindingGeneration })`. The app host downloads artifacts, verifies size/SHA-256, and owns the `ArtifactReader` lifecycle. This contract is unchanged.
2. **Local (CLI, examples, firmware-updater-web):** pass component `ArrayBuffer`s and/or `resourceArchiveBinary`. Core parses the ZIP, compares RESC headers, and writes only changed packages.

Removed `prepareFirmwareUpdateV4MemoryHost`, `FirmwareUpdateV4MemoryHost`, and `buildProtocolV2LocalFirmwareUpdatePlan`. Local installs no longer wrap binaries in a memory `PreparedPlan`.

Other resource-update rules:

- Device write paths come from the signed RESC header (`flexible_metadata`), not a host-side path allowlist.
- Skip a package when the live on-device header already matches. Boot resource is compared against mounted `boot_resource.okpkg`; leftover `.okpkg.staging` is deleted and the package is not transferred again when the live hash matches.
- ZIP uniqueness uses the full archive path so same-named packages from different directories do not collide.
- Invalid / traversal / control-character paths still fail before the first device write.

## Compatibility and risk

- App firmware update in `app-monorepo` continues to use `executePreparedFirmwareUpdateV4` (`preparedPlan` + `hostBindingGeneration`). It does not import MemoryHost.
- `hd-cli firmware-update-v4` already required local binaries / `--resource-archive`; that path is unchanged.
- expo-example, expo-playground, and firmware-updater-web now download Plan artifacts and pass them as local binaries.
- Published package on this branch lineage: `@onekeyfe/hd-core 1.2.0-alpha.159`. MemoryHost removal is not in 157 yet; it needs a later alpha before App should depend on the deletion. App does not need that deletion to keep working.

## Test plan

- Unit tests for BLE name canonicalization and Pro vs Pro2 collisions.
- Protocol V2 resource/firmware tests for header paths, live boot-resource skip, and local ZIP install.
- Example `loadFirmwareUpdatePlanBinaries` tests.
- firmware-updater-web ServiceHardware tests for remote Plan → local binaries.


originalix:

@wabicai

Reviewed commit `302f9ff916b3`. This report always reflects the latest reviewed changes.

### Review summary

The PR moves local Protocol V2 firmware and resource updates from an in-memory PreparedPlan host to direct component buffers and resource ZIP sources in Core, consolidates RESC archive parsing and staging comparisons, canonicalizes Pro2 BLE names, updates example integrations, and aligns package versions.

**What needs attention:** Restore the RESC filesystem boundary, reject incomplete mixed firmware/resource requests before entering the loader, preserve plan identity and version safeguards in example-driven updates, and retain the removed public firmware-host exports during migration.

### Issues to address

- **P1 · Protocol V2 resource packages can write outside supported device paths**
  - **Impact:** The RESC path validator now rejects only malformed or traversal-like strings. FirmwareUpdateV4 writes the header-provided path directly, expanding the loader filesystem write surface beyond the documented resource roots and staging file.
  - **Suggested action:** Restore a positive allowlist for supported resource roots and narrowly enumerated staging paths, rejecting all other header paths before entering loader mode.
- **P1 · Direct resource updates can omit requested firmware targets**
  - **Impact:** When a resource archive is supplied, the early direct-artifact path filters only the binaries that happen to be present and does not verify every requested non-resource target. A mixed request can therefore report success after applying only resources.
  - **Suggested action:** Validate all requested non-resource targets against the available install items before opening sources or entering the loader, and return a preparation error for any missing target.
- **P1 · Plan-driven example updates drop device identity and version safeguards**
  - **Impact:** The new plan adapter returns only binary fields and targets, so example callers invoke firmwareUpdateV4 without the plan's device identity or expected target versions. The update is treated as an ordinary local update, allowing a reconnect to a different device and skipping the plan's post-install version checks.
  - **Suggested action:** Keep plan-driven example updates on the prepared-plan/host-binding path, or propagate equivalent device identity, model, firmware type, and expected-version checks into the direct invocation.
- **P2 · Published firmware-host exports are removed without a compatibility period**
  - **Impact:** The hd-core package barrel no longer exports prepareFirmwareUpdateV4MemoryHost or its related types, so existing updater integrations importing those symbols will fail to compile after upgrading.
  - **Suggested action:** Retain a deprecated compatibility wrapper and the associated type exports for at least one release while consumers migrate to direct binaries or a persistent PreparedPlan host.

### Validation gaps

`git diff --check` passed. The focused Core Jest command could not run because the checkout has no Jest executable available. Physical-device validation is still needed for Pro2/Neo resource-only and mixed firmware/resource updates, boot-resource staging promotion, reconnect identity handling, and stale BLE bond behavior.

packages/core/src/protocols/protocol-v2/resources.ts
```diff
@@ -49,7 +49,15 @@ export function parseProtocolV2Resources(value: unknown): IProtocolV2Resources |
   };
 }
 
-const PROTOCOL_V2_RESOURCE_DEVICE_ROOTS = ['vol0:/bundles/', 'vol0:/loaders/rom/'] as const;
+const PROTOCOL_V2_RESOURCE_DEVICE_ROOTS = [
```
originalix: P1: [Protocol V2 resource path whitelist now accepts unsupported bootloader targets]

This change makes `parseProtocolV2ResourcePackage()` accept any `vol0:/loaders/bootloader/*.okpkg` or `*.okpkg.staging` path.

The current resource-update contract only allows `vol0:/bundles/**/*.okpkg`, `vol0:/loaders/rom/**/*.okpkg`, and the single `vol0:/loaders/bootloader/boot_resource.okpkg.staging` handoff path. Once an archive with one of these newly accepted bootloader paths passes validation, the transfer phase no longer fails early and can write directly into bootloader-owned files or treat an unrelated staging file as boot-resource staging.

Please restore an explicit whitelist for the supported paths here, or update the transfer logic and the documented contract together if firmware really introduced additional bootloader resource targets.

packages/core/src/api/SearchDevices.ts
```diff
@@ -40,7 +40,7 @@ export default class SearchDevices extends BaseMethod {
           seenIds.add(lowerId);
           const rawBleName =
             device.name ?? (device as unknown as { localName?: string }).localName ?? '';
-          const bleName = normalizePro2FindMyAdvertisementName(rawBleName);
+          const bleName = canonicalizePro2BleAdvertisementName(rawBleName);
```
originalix, line 43: P2: [BLE name canonicalization breaks exact name selection for existing callers]

This rewrites the discovered Pro2 name from `Pro2 XXXX` to `Pro 2 XXXX` before it leaves `searchDevices()`.

Callers that still compare names literally, such as the CLI `--device-name` flow, now stop matching existing user input and scripts that pass the legacy compact form. The device is still present, but the lookup fails because the SDK changed the public name without updating the exact-match consumers.

If the spaced form is the new display name, the lookup path needs to compare canonicalized aliases on both sides (for example via `isSameOnekeyBleName`) or preserve the original advertisement string alongside the canonical display name.

packages/core/src/protocols/protocol-v2/resources.ts
```diff
@@ -49,25 +49,15 @@ export function parseProtocolV2Resources(value: unknown): IProtocolV2Resources |
   };
 }
 
-const PROTOCOL_V2_RESOURCE_DEVICE_ROOTS = ['vol0:/bundles/', 'vol0:/loaders/rom/'] as const;
-
-function isAllowedResourceDevicePath(path: string): boolean {
-  if (
+function isSafeResourceDevicePath(path: string): boolean {
```
originalix, line 66: P1: [Protocol V2 resource validation accepts unsupported device paths]

This validator now rejects only traversal-like syntax, so a RESC header such as `vol0:/unexpected/images.okpkg` or an arbitrary bootloader filename passes parsing.

The transfer phase uses the header value directly as the `FilesystemFileWrite` path. That lets an unsupported resource archive reach device I/O and potentially replace bootloader-owned or otherwise unintended files instead of failing before the first write.

Please restore an explicit allowlist for `vol0:/bundles/**/*.okpkg`, `vol0:/loaders/rom/**/*.okpkg`, and only `vol0:/loaders/bootloader/boot_resource.okpkg.staging`, with negative tests for other roots and suffixes.

packages/core/src/api/SearchDevices.ts
originalix, line 43: P2: [BLE name canonicalization breaks legacy exact-name selection]

This changes a discovered `Pro2 XXXX` name to `Pro 2 XXXX` before it leaves `searchDevices()`.

The CLI and other existing callers can still compare the returned name literally, so `--device-name Pro2 XXXX` no longer finds the same device even though discovery succeeds. This breaks existing scripts and user input that rely on the compact form.

Please compare the requested and discovered names through the canonical alias helper on both sides, or preserve the raw advertisement name for lookup while using the spaced form only for display.

packages/core/src/protocols/protocol-v2/resources.ts
```diff
@@ -49,25 +49,29 @@ export function parseProtocolV2Resources(value: unknown): IProtocolV2Resources |
...
+  return (
+    fileName.length > 0 &&
+    !fileName.startsWith('.') &&
+    !parts.some(part => part === '__MACOSX' || part === '.' || part === '..' || part === '')
+  );
+}
 
-function isAllowedResourceDevicePath(path: string): boolean {
-  if (
+function isSafeResourceDevicePath(path: string): boolean {
```
originalix, line 66: P1: [Protocol V2 resource validation accepts unsupported device paths]

This validator now rejects only traversal-like syntax, so a RESC header such as `vol0:/unexpected/images.okpkg` or an arbitrary bootloader filename passes parsing.

The transfer phase uses the header value directly as the `FilesystemFileWrite` path. That lets an unsupported resource archive reach device I/O and potentially replace loader-owned or otherwise unintended files instead of failing before the first write.

Please restore an explicit allowlist for `vol0:/bundles/**/*.okpkg`, `vol0:/loaders/rom/**/*.okpkg`, and `vol0:/loaders/bootloader/**/*.okpkg.staging`, with negative tests for other roots and suffixes.

packages/core/src/api/SearchDevices.ts
originalix, line 43: P2: [BLE name canonicalization breaks legacy exact-name selection]

This changes a discovered `Pro2 XXXX` name to `Pro 2 XXXX` before it leaves `searchDevices()`.

The CLI and other existing callers can still compare the returned name literally, so `--device-name Pro2 XXXX` no longer finds the same device even though discovery succeeds. This breaks existing scripts and user input that rely on the compact form.

Please compare the requested and discovered names through `isSameOnekeyBleName()` at the selection boundary, or preserve the raw advertisement name for lookup while using the spaced form only for display.

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource updates now accept writes outside the documented device roots]

When a RESC package carries any syntactically "safe" path, this helper now accepts it. `prepareProtocolV2LocalResourceArchive()` later trusts `header.devicePath`, and `executeProtocolV2TransferPhase()` writes the package to that path verbatim.

That means a ZIP containing `vol0:/unexpected/...` or another bootloader-side `.okpkg.staging` path is now treated as valid update content instead of being rejected before the first device write. In the firmware/resource update path, widening the allowed filesystem targets is a merge blocker because it can redirect writes outside the documented `bundles` / `loaders/rom` roots and bypass the special-case handling that was only meant for `boot_resource.okpkg.staging`.

Please restore an explicit allowlist here and only broaden it once the device contract and updater flow are updated together.

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
```diff
@@ -71,6 +71,11 @@ export const getFirmwareUpdateResourceName = (value: unknown): string => {
   return resourceName;
 };
 
+const getPreparedEntryIdentity = (entryName: unknown): string => {
```
originalix, line 74: P2: [Shared prepared-plan validation no longer matches V2/V3 resource archive consumers]

This change switches duplicate detection from the canonical resource name to the full ZIP entry path. A plan with two different directories that contain the same basename now passes `prepareFirmwareUpdatePlan()` and `validateFirmwareUpdatePreparedPlan()`.

The problem is that V2 and V3 still consume prepared resource ZIPs through `readVerifiedPreparedResourceArchive()`, which canonicalizes each entry back to the basename and rejects the same archive as duplicate. So the shared validator can now approve a prepared plan that those executors are guaranteed to fail at runtime.

Please keep the shared validator aligned with its consumers: either preserve basename uniqueness for shared prepared plans, or scope the path-based rule to the V4 resource flow and update the V2/V3 archive reader in the same change.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/shared/src/constants.ts
```diff
@@ -292,6 +292,30 @@ export const normalizePro2FindMyAdvertisementName = (value: string) => {
...
 };
 
+/**
+ * Current Pro2 advertisements use "Pro 2 XXXX". Older firmware used "Pro2 XXXX".
+ * Discovery and DeviceInfo both go through this helper so the public BLE name
+ * stays on the spaced form without changing OneKey Pro / Neo names.
+ */
+export const canonicalizePro2BleAdvertisementName = (value: string) => {
+  const withoutFindMy = normalizePro2FindMyAdvertisementName(value);
+  const match = withoutFindMy.match(/^(onekey\s*)?pro\s*2\s*/i);
```
sidmorizon, line 323: # 🟠 P1 一代 OneKey Pro 名称被规范化逻辑改写

## 问题

`/^(onekey\s*)?pro\s*2\s*/i` 只要求 `Pro` 后面紧跟一个 `2`，并不检查这个 `2` 是不是型号位。一代 OneKey Pro 的广播名是 `Pro XXXX`（4 位十六进制后缀），当后缀首字符恰好是 `2` 时，该正则会把序号的第一位当成型号吃掉：

```
canonicalizePro2BleAdvertisementName('Pro 22D8')        // => 'Pro 2 2D8'
canonicalizePro2BleAdvertisementName('Pro 2D8F')        // => 'Pro 2 D8F'
canonicalizePro2BleAdvertisementName('OneKey Pro 22D8') // => 'OneKey Pro 2 2D8'
```

现有单测只覆盖了 `Pro A1B2` 这种后缀以字母开头的样例，所以这条路径没有被挡住。

## 影响

后缀首位为 `2` 的一代 Pro 约占 `1/16`。这些设备经 `SearchDevices`（`name`/`deviceType`）、`resolveDeviceBleName`、`buildProtocolV2FeaturesPayload`、`Device.getCurrentBleName()` / `getCurrentDisplayName()` 输出后，公开的 BLE 名称会变成设备实际根本没有广播的 `Pro 2 2D8`；`getDeviceTypeByBleName('Pro 2 2D8')` 也会稳定命中 `\bPro\s*2\b` 判成 `Pro2`。结果是 UI 显示错误名称、按名称精确匹配（如 CLI `--device-name`）失配、设备型号被误判为二代。

## 建议

规范化前先确认整名确实是 Pro2 命名，例如复用已有的 `PRO2_COMPACT_NAME_PATTERN` 对 `compactBleName(value)` 做校验后再重写，并补上 `Pro 22D8` / `Pro 2D8F` / `OneKey Pro 22D8` 保持原样的负向用例：

```ts
const compact = compactBleName(withoutFindMy);
if (!PRO2_COMPACT_NAME_PATTERN.test(compact)) return withoutFindMy;
```

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

sidmorizon (commented):
Review

---

packages/shared/src/constants.ts
```diff
@@ -292,6 +292,30 @@ export const normalizePro2FindMyAdvertisementName = (value: string) => {
...
+  if (!match) return withoutFindMy;
+
+  const rest = withoutFindMy.slice(match[0].length);
+  const prefix = match[1] ? 'OneKey Pro 2' : 'Pro 2';
+  return rest ? `${prefix} ${rest}` : prefix;
+};
```
sidmorizon, line 329: 🟠 P1: Some OneKey Pro devices get renamed and misidentified as Pro 2 during Bluetooth discovery

Severity: `severe`

A OneKey Pro whose Bluetooth name suffix begins with the digit 2 is rewritten into a Pro 2 name (`canonicalizePro2BleAdvertisementName` at `packages/shared/src/constants.ts:300-308`), so those Pro devices are shown with a wrong name and reported as a different product.
Impact: Roughly one in sixteen OneKey Pro devices appears in the device list as "Pro 2 …" and is classified as a Pro 2, which can drive wrong product-specific behavior such as firmware/capability lookups.

<details>
<summary>Prefix regex consumes the first digit of a Pro serial suffix</summary>

The prefix pattern `/^(onekey\s*)?pro\s*2\s*/i` treats the leading `2` of a 4-hex suffix as the model token: `"Pro 22D8"` → `"Pro 2 2D8"`, `"Pro 2E5F"` → `"Pro 2 E5F"` (verified by evaluating the regex). `"Pro A1B2"` is unaffected, which is why the added test at `packages/core/__tests__/device-utils.test.ts:137` passes.

Because the canonicalizer is now applied on every BLE-name path — discovery (`packages/core/src/api/SearchDevices.ts:43`), Protocol V1/V2 feature building (`packages/core/src/utils/deviceFeaturesCompat.ts:135`, `packages/core/src/deviceProfile/buildDeviceFeatures.ts:253`, `packages/core/src/device/DeviceStateMapper.ts:251`) and display (`packages/core/src/device/Device.ts:676-677`) — the mangled name is what users see and what `getDeviceTypeByBleName` classifies. `getDeviceTypeByBleName` correctly leaves `"Pro 22D8"` as `EDeviceType.Pro` (`/\bPro\s*2\b/` fails on `Pro 22D8`), but after canonicalization the input becomes `"Pro 2 2D8"`, which matches and yields `EDeviceType.Pro2` (`packages/core/src/utils/deviceInfoUtils.ts:33-36`).

`isSameOnekeyBleName` inherits the same defect, so a stored `"Pro 22D8"` and a freshly advertised `"Pro 22D8"` still compare equal, but a Pro name would be compacted to `pro22d8` and could collide with the compact Pro 2 form `Pro2 2D8`.

</details>

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource updates accept writes outside the documented device roots]

This helper now accepts any RESC header path that avoids traversal and control characters. `FirmwareUpdateV4` later trusts `header.devicePath` and writes the resource to that path verbatim, while the staging check treats every `.okpkg.staging` file under the bootloader directory as the special boot-resource file.

A malformed or compromised archive can therefore redirect writes outside the supported resource roots and bypass the exact staging-file contract. Please restore an explicit allowlist and keep only the documented boot-resource staging path as the special case.

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
```diff
@@ -71,6 +71,13 @@ export const getFirmwareUpdateResourceName = (value: unknown): string => {
...
```
originalix, line 74: P2: [Prepared-plan path uniqueness diverges from V2/V3 resource archive validation]

This change makes duplicate detection use the full ZIP path, so entries such as `loaders/rom/params.okpkg` and `loaders/bootloader/params.okpkg` can both pass plan validation.

V2 and V3 still read prepared resource archives through `FirmwarePreparedResourceArchive`, which canonicalizes entries to their basenames and rejects that same archive as a duplicate. The host can therefore approve a plan that execution will fail to consume. Please keep shared validation basename-compatible, or update the reader and all affected executors together.

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource updates can write outside approved device roots]

When a local or prepared resource ZIP contains a printable, non-traversing RESC path such as `vol0:/unexpected/images.okpkg`, this helper accepts it. `FirmwareUpdateV4` then passes `header.devicePath` directly to `FilesystemFileWrite`, so package bytes can overwrite content outside the supported resource locations.

Please restore a root-and-suffix allowlist for approved bundle and ROM `.okpkg` destinations and bootloader `.okpkg.staging` destinations before trusting the header path.

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
originalix, line 74: P2: [Prepared-plan uniqueness no longer matches V2 and V3 archive consumption]

When a ZIP contains the same basename in different directories, this full-path identity lets the prepared plan pass validation. FirmwareUpdateV2 and FirmwareUpdateV3 later use `FirmwarePreparedResourceArchive`, which reduces those entries to basenames and rejects the same archive as a duplicate, so an approved plan fails during execution.

Please make this identity role-aware: keep basename uniqueness for legacy `resource` artifacts and use full normalized paths only for V4 `resourceBundle` artifacts.

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource updates can write outside approved device roots]

When a local or prepared resource ZIP contains a printable, non-traversing RESC path such as `vol0:/unexpected/images.okpkg`, this helper accepts it. `FirmwareUpdateV4` then passes `header.devicePath` directly to `FilesystemFileWrite`, so package bytes can overwrite content outside the supported resource locations.

Please restore a root-and-suffix allowlist for approved bundle and ROM `.okpkg` destinations and bootloader `.okpkg.staging` destinations before trusting the header path.

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
originalix, line 74: P2: [Prepared-plan uniqueness no longer matches V2 and V3 archive consumption]

When a ZIP contains the same basename in different directories, this full-path identity lets the prepared plan pass validation. FirmwareUpdateV2 and FirmwareUpdateV3 later use `FirmwarePreparedResourceArchive`, which reduces those entries to basenames and rejects the same archive as a duplicate, so an approved plan fails during execution.

Please make this identity role-aware: keep basename uniqueness for legacy `resource` artifacts and use full normalized paths only for V4 `resourceBundle` artifacts.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -792,15 +780,7 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
-      const localMemoryHost = await this.prepareProtocolV2LocalMemoryHost({
-        features: deviceFeatures,
-        firmwareType,
-      });
-      try {
-        return await this.runProtocolV2PreparedArtifacts(deviceFeatures, firmwareType);
-      } finally {
-        localMemoryHost.release();
-      }
+      return this.runProtocolV2DirectArtifacts({});
```
sidmorizon, line 783: # 🟠 P1 本地资源直装路径丢失「请求的固件目标缺少二进制」校验

## 问题

本次提交把这个分支从 `prepareProtocolV2LocalMemoryHost()` + `runProtocolV2PreparedArtifacts()` 换成了 `runProtocolV2DirectArtifacts({})`，但 `prepareProtocolV2LocalMemoryHost()` 里被一并删掉的这段前置校验没有在新路径上补回：

```ts
const missingTarget = Array.from(requestedComponentTargets).find(
  target => !localComponentTargets.has(target)
);
if (missingTarget) {
  throw ERRORS.TypedError(HardwareErrorCode.RuntimeError,
    `Protocol V2 local update has no binary for requested target ${missingTarget}`,
    { firmwareUpdateCode: 'FirmwareArtifactsNotPrepared', artifactName: missingTarget });
}
```

新路径里 `runProtocolV2DirectArtifacts()` 只做 `filterProtocolV2LocalInstallItems()` 过滤（保留交集），没有任何「请求了但没有对应二进制」的反向检查。这个分支在 `runProtocolV2()` 最前面就 `return`，所以后面的 `getMissingProtocolV2FirmwareTargets()` / `prepareRemoteProtocolV2Binaries()` 的缺失目标兜底也都不会执行。

## 影响

调用方传 `targetsToUpdate: ['resource', 'app_v1']` + `resourceArchiveBinary`，但忘记（或上层构造失败导致）没有传 `applicationP1Binary` 时：旧行为是抛 `FirmwareArtifactsNotPrepared`；新行为是 `installItems` 过滤后为空、`resourceSources` 非空，于是只装资源包，`executeProtocolV2SourceUpdate()` 正常走完 `completeProtocolV2FinalVerification()` 并返回成功。

`assertExpectedProtocolV2Versions()` 只有在调用方传了 `expectedTargetVersions` 时才会兜住，缺省情况下没有任何目标覆盖度断言。结果是固件更新静默少装一个组件却上报成功，设备停留在旧版本，属于固件写入路径上的静默 under-install。另外这个分支同样绕过了 `runProtocolV2DirectArtifacts` 之外的 `external-only` manifest 守卫，缺失校验后错误反馈完全消失。

## 建议

在 `runProtocolV2DirectArtifacts()` 里对过滤后的 `installItems` 复用现成的 `getMissingProtocolV2FirmwareTargets()` 做一次 fail-closed 检查，保持原来的 `FirmwareArtifactsNotPrepared` 错误码与 `artifactName`：

```ts
const installItems = this.filterProtocolV2LocalInstallItems(...);
const missing = this.getMissingProtocolV2FirmwareTargets(installItems);
if (missing.length) {
  throw ERRORS.TypedError(
    HardwareErrorCode.RuntimeError,
    `Protocol V2 local update has no binary for requested target ${missing[0]}`,
    { firmwareUpdateCode: 'FirmwareArtifactsNotPrepared', artifactName: missing[0] }
  );
}
```

并补一条「本地 ZIP + 请求 app_v1 但未提供 applicationP1Binary 时抛错」的用例——当前 `protocol-v2.test.ts` 里没有任何测试覆盖这条被删除的分支。

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

sidmorizon (commented):
增量 review：`0f330a9` 的资源直装改造。1 处 P1。

---

packages/connect-examples/expo-example/src/views/FirmwareScreen/firmwareUpdatePlanHost.ts
```diff
@@ -101,39 +77,49 @@ async function loadPlanArtifact({
...
+  const loadedTargets: FirmwareUpdateV4Target[] = [];
+  for (const item of loaded) {
+    if (binaries[item.field]) {
+      throw new Error(`Firmware update Plan has duplicate target ${item.target}`);
+    }
+    binaries[item.field] = item.binary;
+    loadedTargets.push(item.target);
+  }
+
+  binaries.targetsToUpdate = [...new Set(loadedTargets)];
```
sidmorizon, line 123: # 🟡 P2 Plan 驱动的本地安装丢失安装后版本断言

## 问题

迁移前 `prepareFirmwareUpdatePlanMemoryHost()` 把整个 `plan` 交给 `prepareFirmwareUpdateV4MemoryHost()`，Core 在构造 `this.params` 时会从 `preparedPlan.artifacts[].targetVersion` 反推出 `expectedTargetVersions`（`packages/core/src/api/FirmwareUpdateV4.ts:671-685` 与 `:718-720`）。迁移后 `loadFirmwareUpdatePlanBinaries()` 只返回二进制字段和 `targetsToUpdate`，`plan.artifacts[].targetVersion`（由 `FirmwareUpdatePlan.ts:542/807/831/877` 从 release 填入）被整体丢弃，调用方 `index.tsx` 也没有补传：

```ts
const binaries = await loadFirmwareUpdatePlanBinaries({ plan });
const res = await sdk.firmwareUpdateV4(selectDevice.connectId, {
  platform: params.platform,
  ...binaries, // 只有 targetsToUpdate + 各组件 ArrayBuffer
});
```

## 影响

`assertExpectedProtocolV2Versions()` 首行是 `if (!expected) return;`，因此 `completeProtocolV2FinalVerification()`（`FirmwareUpdateV4.ts:2650`）里的安装后版本断言在这两个宿主上完全空转。设备重启后若某个组件仍停留在旧版本（写入被跳过、staging 未提升、SE 未真正激活），`firmwareUpdateV4` 依然返回 `success`，之前会抛错的静默 under-install 现在无人兜底。`packages/connect-examples/expo-playground/app/utils/firmwareUpdatePlanHost.ts` 是同一份拷贝，问题一致。

## 建议

把 `expectedTargetVersions` 纳入 `FirmwareUpdatePlanBinaryParams`，在遍历 artifact 时顺带收集，保持与旧 PreparedPlan 路径等价的收尾校验：

```ts
const expectedTargetVersions: NonNullable<
  FirmwareUpdateV4Params['expectedTargetVersions']
> = {};
for (const item of loaded) {
  if (item.target !== 'resource' && item.targetVersion) {
    expectedTargetVersions[item.target] = item.targetVersion;
  }
}
```

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

sidmorizon (commented):
增量 review：`0cef8e1`（移除 firmwareUpdateV4 memory host）。1 条 P2。

---

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource packages can now write to arbitrary device paths]

When a RESC package declares a header path like `vol0:/unexpected/foo.okpkg`, this parser now accepts it as long as the string is only traversal-safe. `FirmwareUpdateV4` later writes to `header.devicePath` verbatim, so the trusted write surface has expanded from the documented resource roots to any syntactically valid file path on the device.

That means a mispackaged release archive, or a locally supplied archive, can overwrite files outside the supported bundles / ROM / boot-resource locations. Please restore an explicit allowlist here, or add a narrow allowlist for any newly supported paths before we trust the header metadata.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1055,133 +1028,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
-          firmwareUpdateCode: 'FirmwareArtifactsNotPrepared',
-          artifactName: missingTarget,
-        }
+    announceDownload?: boolean;
+  }) {
+    try {
+      if (announceDownload) {
+        this.postTipMessage(FirmwareUpdateTipMessage.StartDownloadFirmware);
+      }
+      const installItems = this.filterProtocolV2LocalInstallItems(
```
originalix, line 1042: P1: [Direct local V4 updates no longer reject missing requested firmware targets]

When `targetsToUpdate` includes both firmware and `resource`, this path now filters whatever local binaries happen to exist and proceeds as long as there is at least one install source or resource bundle. The previous local-memory-host flow rejected the request if any non-resource target was missing.

That means a call such as `targetsToUpdate: ['app_v1', 'resource']` plus only `resourceArchiveBinary` will now succeed as a resource-only update, leaving firmware and resources out of sync while the caller believes the full target set was applied.

Please reintroduce the missing-target check here by calling `getMissingProtocolV2FirmwareTargets()` and failing with `FirmwareArtifactsNotPrepared` before transfer when any requested component binary is absent.

packages/core/src/index.ts
```diff
@@ -21,12 +21,6 @@ export { executeCallback, cleanupCallback };
 export { preloadSessionCache } from './device/Device';
 export { projectFeatures as projectDeviceStateFeatures } from './device/DeviceStateProjector';
 export { getMethodSupportedProtocols } from './api/utils';
-export { prepareFirmwareUpdateV4MemoryHost } from './api/firmware/FirmwareMemoryHost';
-export type {
-  FirmwareMemoryArtifact,
-  FirmwareMemoryArtifactEntry,
-  FirmwareUpdateV4MemoryHost,
-} from './api/firmware/FirmwareMemoryHost';
 export {
```
originalix, line 24: P2: [`@onekeyfe/hd-core` drops exported firmware-host helpers without compatibility handling]

This barrel no longer re-exports `prepareFirmwareUpdateV4MemoryHost` and the related memory-host types, and the implementation file is deleted in the same change. Any external updater host importing those symbols from `@onekeyfe/hd-core` will start failing to compile as soon as it picks up this alpha.157 release.

If the API really has to go away, please keep a deprecated shim for one release or document and version this as an explicit breaking change with a migration path.

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource packages can now write to arbitrary device paths]

A RESC package that declares a path such as `vol0:/unexpected/foo.okpkg` now passes validation as long as it avoids traversal and control characters. `FirmwareUpdateV4` then writes to `header.devicePath` directly, so a locally supplied or mispackaged archive can target files outside the supported resource and boot-resource locations.

Please restore an explicit allowlist here, including only narrowly defined newly supported paths, before trusting the header metadata.

packages/core/src/api/FirmwareUpdateV4.ts
originalix, line 1042: P1: [Direct local V4 updates no longer reject missing requested firmware targets]

When `targetsToUpdate` includes both firmware and `resource`, this path filters only the binaries that were supplied and proceeds as soon as a resource archive exists. A call such as `targetsToUpdate: ['app_v1', 'resource']` with only `resourceArchiveBinary` therefore succeeds as a resource-only update.

Please call `getMissingProtocolV2FirmwareTargets()` (or an equivalent check) before transfer and return `FirmwareArtifactsNotPrepared` whenever a requested non-resource target has no local binary.

packages/core/src/index.ts
originalix, line 24: P2: [hd-core removes exported firmware-host helpers without compatibility handling]

This barrel no longer exports `prepareFirmwareUpdateV4MemoryHost` or the related memory-host types, and the implementation file is deleted in the same change. External updater hosts importing these published symbols from `@onekeyfe/hd-core` will fail to compile on upgrade.

Please keep a deprecated compatibility wrapper for one release, or document and version this as an intentional breaking change with a migration path.

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: tmp

packages/core/src/api/SearchDevices.ts
wabicai: This is a duplicate of PRRT_kwDOHV2MTM6aaRNP and has been consolidated there; resolving to reduce noise.

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: resource path 风险问题重复条目，已与 PRRT_kwDOHV2MTM6agDOF 合并。

wabicai: resource path 问题重复条目，统一到 PRRT_kwDOHV2MTM6agDOF 跟踪。

wabicai: 与 PRRT_kwDOHV2MTM6agDOF 为同一条 resource path 风险问题的重复反馈，已合并到 agDOF。

wabicai: resource path 风险问题重复条目，保留 agDOF 主讨论。

packages/core/src/api/SearchDevices.ts
wabicai: 此条与 PRRT_kwDOHV2MTM6agDSl/aaRNP 重复，已合并到 PRRT_kwDOHV2MTM6aaRNP 跟踪。

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: 与 PRRT_kwDOHV2MTM6agDOF 属于同一风险点（resource 路径收口），已合并留一个主讨论。

packages/core/src/index.ts
wabicai: 与 PRRT_kwDOHV2MTM6ajqdC 重复，保持后一条跟踪导出兼容性。

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
wabicai: 与 PRRT_kwDOHV2MTM6ahCJd duplicate。

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: 与 PRRT_kwDOHV2MTM6ajqZe 重复，合并到同一主线。

packages/core/src/api/firmware/FirmwareUpdatePreparedPlan.ts
wabicai: 与 PRRT_kwDOHV2MTM6ahCJd 为同一 prepared-plan 重复检测问题，已合并。

wabicai: 与 PRRT_kwDOHV2MTM6ahCJd 重复，合并讨论。

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: 与 PRRT_kwDOHV2MTM6agDOF 为同一资源路径风险问题，已合并。

wabicai: 与 PRRT_kwDOHV2MTM6agDOF 重复，保留后者统一跟进。

wabicai:
我已把明显重复/冗余线程 resolve；当前保留未处理: PRRT_kwDOHV2MTM6aaRNP, PRRT_kwDOHV2MTM6agDOF, PRRT_kwDOHV2MTM6ahCJd, PRRT_kwDOHV2MTM6ahNM8, PRRT_kwDOHV2MTM6ajaha, PRRT_kwDOHV2MTM6ajnYr, PRRT_kwDOHV2MTM6ajqZe, PRRT_kwDOHV2MTM6ajqdC。已resolve 其余重复项，仅做沟通归并。

wabicai:
已按你反馈核对两条 PR 的 review threads，hardware-js-sdk#890 当前全部 review thread 已 resolve（含已修复/无需跟进项）。

packages/shared/src/constants.ts
wabicai: Fixed. The rewrite now gates on the compact Pro2 pattern `^(?:onekey)?pro2[a-f0-9]{4}$` before the `pro\s*2` prefix regex, so a OneKey Pro suffix that starts with `2` is no longer treated as a model token.

- `Pro 22D8` / `Pro 2D8F` / `OneKey Pro 22D8` stay unchanged and classify as Pro
- `Pro 2 A1B2` / `Pro2A1B2` still canonicalize to `Pro 2 A1B2` and classify as Pro2
- `getDeviceTypeByBleName` no longer uses a bare `^Pro2` prefix; it requires `\bPro\s*2\b` or compact `Pro2` + 4 hex
- Covered in `packages/shared/src/constants.test.ts` and `packages/core/__tests__/device-utils.test.ts`

Resolving.

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource packages can now write to arbitrary device paths]

A RESC package that declares a path such as `vol0:/unexpected/foo.okpkg` now passes validation as long as it avoids traversal and control characters. `FirmwareUpdateV4` then writes to `header.devicePath` directly, so a locally supplied or mispackaged archive can target files outside the supported resource and boot-resource locations.

Please restore an explicit allowlist here, including only narrowly defined newly supported paths, before trusting the header metadata.

packages/core/src/api/FirmwareUpdateV4.ts
originalix, line 1042: P1: [Direct local V4 updates no longer reject missing requested firmware targets]

When `targetsToUpdate` includes both firmware and `resource`, this path filters only the binaries that were supplied and proceeds as soon as a resource archive exists. A call such as `targetsToUpdate: ['app_v1', 'resource']` with only `resourceArchiveBinary` therefore succeeds as a resource-only update.

Please call `getMissingProtocolV2FirmwareTargets()` (or an equivalent check) before transfer and return `FirmwareArtifactsNotPrepared` whenever a requested non-resource target has no local binary.

packages/core/src/index.ts
originalix, line 24: P2: [hd-core removes exported firmware-host helpers without compatibility handling]

This barrel no longer exports `prepareFirmwareUpdateV4MemoryHost` or the related memory-host types, and the implementation file is deleted in the same change. External updater hosts importing these published symbols from `@onekeyfe/hd-core` will fail to compile on upgrade.

Please keep a deprecated compatibility wrapper for one release, or document and version this as an intentional breaking change with a migration path.

packages/hd-transport-react-native/src/bleStaleBond.ts
```diff
@@ -0,0 +1,63 @@
...
+
+export const isNativeBleStaleBondError = (error: unknown): boolean => {
+  if (!error || typeof error !== 'object') {
+    return typeof error === 'string' ? isBleStaleBondErrorText(error) : false;
+  }
+
+  const nativeError = error as NativeBleErrorFields;
+  if (
+    nativeError.attErrorCode === ATT_INSUFFICIENT_AUTHENTICATION ||
+    nativeError.attErrorCode === ATT_UNLIKELY_ERROR ||
```
sidmorizon: # 🟠 P1 ATT `0x0E` 被当作陈旧配对，会误导用户删除配对

## 问题

`ATT_UNLIKELY_ERROR = 14` 即 ATT `0x0E` Unlikely Error，蓝牙规范里它是"请求遇到了不太可能发生的错误、无法完成"的通用兜底码，和加密/鉴权无关；真正表示陈旧绑定的是 `0x05` Insufficient Authentication 与 `0x0F` Insufficient Encryption，这两个已经分别由 `ATT_INSUFFICIENT_AUTHENTICATION` 和 `ATT_INSUFFICIENT_ENCRYPTION` 覆盖。

改动前 `remapError()` 判断的是 `error.iosErrorCode === BleATTErrorCode.UnlikelyError`，字段是 `iosErrorCode`，数值 `14` 恰好等于 `CBError.peerRemovedPairingInformation`，语义是对的（常量名用错了而已）。本次重构把它拆成了两个条件，`iosErrorCode === 14` 保留了原语义，但新增的 `attErrorCode === 14` 是一个此前不存在、且语义完全不同的判断，并且它在 iOS 和 Android 上都会生效（Android GATT status `14` 同样不是绑定类错误）。

```ts
nativeError.attErrorCode === ATT_UNLIKELY_ERROR ||
```

## 影响

`toBleStaleBondHardwareError()` 第 54 行同样用 `attErrorCode === ATT_UNLIKELY_ERROR` 判定 `peerRemoved`，因此一个来自固件的通用 ATT `0x0E` 会被映射成 `HardwareErrorCode.BlePeerRemovedPairingInformation`，其文案是 `need to delete pairing information`，App 会据此引导用户到系统蓝牙里"忽略此设备"并重新配对——这是一个可恢复错误触发的用户可见破坏性流程。

同时该错误会被 `rememberStaleBondError()` 粘在 `staleBondErrors` 上，使本次连接内后续 `detectProtocol()` 直接抛出；而 `packages/core/src/core/index.ts` 的 `ensureConnected` 已把 `BlePeerRemovedPairingInformation` 列入不重试集合（`device-lifecycle-events.test.ts` 断言为 `false`）。结果是一次瞬时固件错误既拿不到自动重试，又把用户推向重新配对。

## 建议

移除 `attErrorCode === ATT_UNLIKELY_ERROR` 这一条，只保留 `0x05`/`0x0F` 以及 `iosErrorCode === 14`（建议把常量改名为 `CB_ERROR_PEER_REMOVED_PAIRING_INFORMATION` 以免和 ATT 枚举混淆）。若线上确实观测到 iOS 在陈旧绑定时只报 ATT `0x0E`，也应额外要求 `iosErrorCode` 或错误文本同时命中，而不是单凭 `attErrorCode` 一个通用码定性。

```ts
const CB_ERROR_PEER_REMOVED_PAIRING_INFORMATION = 14; // CBError, not ATT
// attErrorCode 只接受 0x05 / 0x0F
```

另外 `bleStaleBond.test.ts` 目前只有 `Device disconnected` 一个反例，建议补一个"必须不被判定为陈旧绑定"的 ATT 码用例（例如 `attErrorCode: 3` / `attErrorCode: 6`），锁住分类边界。

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -159,7 +161,30 @@ export default class ElectronBleTransport {
...
 
+  private toStaleBondError(error: unknown): Error | null {
+    if (isBleStaleBondHardwareError(error)) {
+      return error as Error;
+    }
+    const errorMessage =
+      error && typeof error === 'object' && 'message' in error
+        ? String((error as { message?: unknown }).message ?? '')
+        : String(error ?? '');
+    if (!isBleStaleBondErrorText(errorMessage)) {
```
sidmorizon: # 🟡 P2 Electron 侧只靠英文错误文案识别陈旧绑定

## 问题

RN 侧的 `isNativeBleStaleBondError()` 先看结构化字段 `attErrorCode` / `iosErrorCode`，文本匹配只是兜底。而 Electron 这条路径里 `toStaleBondError()` 唯一的识别手段就是 `isBleStaleBondErrorText(errorMessage)`，即对 `error.message` 做 `encryption is insufficient` / `insufficient encryption` / `peer removed pairing information` / `gatt_insuf_authentication` 的英文子串匹配。

这些字符串的来源并不稳定：`Encryption is insufficient.` 是 macOS CoreBluetooth `CBATTErrorDomain` 的 `NSLocalizedDescription`，会随系统语言本地化；Linux BlueZ 与 Windows WinRT 经 noble 抛出的加密/配对失败根本不是这套措辞。而且错误还要跨 `window.desktopApi.nobleBle` 的 IPC 边界，结构化字段通常在序列化中丢失，只剩 `message`。

## 影响

在非英文 macOS，或在 Linux / Windows 的 Electron 上，`subscribe()` 因陈旧绑定失败时 `toStaleBondError()` 返回 `null`，`handleBluetoothError()` 原样抛出原始错误，本 PR 的 fail-fast 静默失效，退回到等待 Ping/GetFeatures 超时的旧行为，且不会有任何日志提示分类漏掉了。这是"悄悄不生效"而不是报错，最难在真机上发现。

另外第 176 行的 `errorMessage.includes('Peer removed pairing information')` 是大小写敏感的，而 `isBleStaleBondErrorText()` 内部做过 `toLowerCase()`；两处判定口径不一致时会把 peer-removed 降级成 `BleDeviceBondError`。

## 建议

让主进程的 noble 处理器把结构化信息透传过 IPC（例如附带 `attErrorCode` / 原始 `code`），Electron 侧优先按码判定、文本仅作兜底，与 RN 侧保持同一套分类逻辑。同时把第 176 行的大小写敏感判断改为复用统一的小写比较，避免两处口径漂移。

```ts
const code = (error as { attErrorCode?: number })?.attErrorCode;
if (code === 0x05 || code === 0x0f) return ERRORS.TypedError(HardwareErrorCode.BleDeviceBondError, errorMessage);
```

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

sidmorizon (commented):
增量复审：BLE 陈旧绑定 fail-fast 提交的错误分类与平台适配。

---

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1055,133 +1028,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
+        availableInstallItems ??
+          this.buildProtocolV2InstallItems({
+            bootloaderBinary: this.prepareBootloaderBinary(),
+            fwBinaryMap: this.collectExplicitTargetBinaries(),
+          })
       );
```
sidmorizon, line 1048: **Requested firmware components can be silently skipped while the update still reports success**

Severity: `non-severe`

Requested firmware parts are silently dropped when the caller supplies a resource archive but no matching component file (`filterProtocolV2LocalInstallItems` at `packages/core/src/api/FirmwareUpdateV4.ts:1042-1048`), so the update finishes as successful while those parts were never written.
Impact: A caller can be told the firmware update succeeded even though components it explicitly asked to update were never installed.

<details>
<summary>Lost missing-target validation in the new direct local artifact path</summary>

Before this PR the local resource path went through `prepareProtocolV2LocalMemoryHost`, which compared `params.targetsToUpdate` against the locally available install items and threw `Protocol V2 local update has no binary for requested target <t>` (`FirmwareArtifactsNotPrepared`) when one was missing.

`runProtocolV2DirectArtifacts` replaces it and only filters: with `firmwareUpdateV4({ targetsToUpdate: ['app_v1','resource'], resourceArchiveBinary })` and no `applicationP1Binary`, `run()` takes the first branch at `packages/core/src/api/FirmwareUpdateV4.ts:783` and reaches `runProtocolV2DirectArtifacts({})`. `buildProtocolV2InstallItems` yields no items, `installSources` is empty, `resourceSources` is non-empty, so the `No firmware to update` guard does not fire and only resources are written, returning success.

`getMissingProtocolV2FirmwareTargets` (`packages/core/src/api/FirmwareUpdateV4.ts:1469-1482`) already computes exactly the missing set and could be reused here.

</details>

packages/core/src/device/DeviceCommands.ts
```diff
@@ -325,7 +325,10 @@ export class DeviceCommands {
         errorCode: error?.errorCode,
         response: getSafeTransportLogPayload(error?.response?.data, type),
       });
-      if (error.errorCode === HardwareErrorCode.BleDeviceBondError) {
+      if (
+        error.errorCode === HardwareErrorCode.BleDeviceBondError ||
+        error.errorCode === HardwareErrorCode.BlePeerRemovedPairingInformation
+      ) {
```
sidmorizon: **Bluetooth pairing-removed failures are reported as a generic bond error**

Severity: `non-severe`

A device whose pairing information was removed is reported with the generic bond error code (the synthetic `BleDeviceBondError` response at `packages/core/src/device/DeviceCommands.ts:328-331`), so the app cannot tell the user their phone/computer pairing must be re-created.
Impact: Users hitting a removed-pairing failure get the wrong recovery guidance, and the newly added distinct error code never reaches callers.

<details>
<summary>The new BlePeerRemovedPairingInformation branch in typedCall is unreachable</summary>

`call()` now folds both `HardwareErrorCode.BleDeviceBondError` and `HardwareErrorCode.BlePeerRemovedPairingInformation` into one synthetic response whose `type` is always `'BleDeviceBondError'` (`packages/core/src/device/DeviceCommands.ts:328-336`). `typedCall` then fails `assertType`, whose message is `assertType: Response of unexpected type: BleDeviceBondError...` (`packages/core/src/device/DeviceCommands.ts:93-99`), so the pre-existing `error.message.indexOf('BleDeviceBondError') > -1` check matches first and throws `BleDeviceBondError`. The newly added `indexOf('BlePeerRemovedPairingInformation')` check at `packages/core/src/device/DeviceCommands.ts:427-429` can never match, because the synthetic type never carries that name and the transport-side message is human text such as `Peer removed pairing information`.

This contradicts the rest of the PR, which deliberately splits the two codes in the React Native transport (`packages/hd-transport-react-native/src/bleStaleBond.ts:45-63`) and adds `BlePeerRemovedPairingInformation` to the Core error lists.

</details>

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1831,6 +1725,12 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     return `vol0:/${path}`;
   }
 
+  private getProtocolV2ResourceComparePath(devicePath: string) {
+    return isProtocolV2BootResourcePackagePath(devicePath)
+      ? PROTOCOL_V2_BOOT_RESOURCE_PACKAGE_PATH
+      : devicePath;
+  }
```
sidmorizon, line 1732: **Non-boot staging resource packages are re-uploaded on every update**

Reported at `packages/core/src/api/FirmwareUpdateV4.ts:1728-1732` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

Only the boot resource is compared against its live installed copy (`getProtocolV2ResourceComparePath` at `packages/core/src/api/FirmwareUpdateV4.ts:1728-1732`), so every other package that writes to a `.staging` path never matches anything on the device and is transferred again on each update.
Impact: Updates repeatedly re-send resource packages that are already installed, making Bluetooth/USB updates noticeably slower than necessary.

<details>
<summary>Comparison path generalization stops at boot_resource</summary>

`isProtocolV2BootResourcePackagePath` matches only the exact path `vol0:/loaders/bootloader/boot_resource.okpkg.staging` (`packages/core/src/api/FirmwareUpdateV4.ts:292-294`). For any other staging package the compare path stays the `.staging` path itself, and `isProtocolV2ResourceBundleUpToDate` reads that path (`packages/core/src/api/FirmwareUpdateV4.ts:1795-1799`). After the previous update, early boot consumed the staging file, so `FilesystemPathInfoQuery` reports it missing, the header read returns `null`, and the bundle is always considered out of date.

This PR's own tests treat such packages as real (`vol0:/loaders/bootloader/params.okpkg.staging` in `packages/core/__tests__/protocol-v2-resources.test.ts:142-146`), and the new case named "skips bootloader params staging when the on-device package is already current" mocks `isProtocolV2ResourceBundleUpToDate` to `true`, so it does not actually exercise the compare path and hides the gap.

Deriving the compare path by stripping a trailing `.staging` from any resource device path would cover all staging packages uniformly.

</details>

Suggested fix:

```ts
  private getProtocolV2ResourceComparePath(devicePath: string) {
    // Staging files are consumed by early boot, so compare against the live
    // package path for every staged resource, not only boot_resource.
    return devicePath.toLowerCase().endsWith('.staging')
      ? devicePath.slice(0, -'.staging'.length)
      : devicePath;
  }
```

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: 可以忽略，这里硬件会进行校验。

packages/core/src/index.ts
wabicai: 忽略，这里会同步修改

packages/core/src/protocols/protocol-v2/resources.ts
originalix, line 66: P1: [Protocol V2 resource packages can now write to arbitrary device paths]

A RESC package that declares a path such as `vol0:/unexpected/foo.okpkg` now passes validation as long as it avoids traversal and control characters. `FirmwareUpdateV4` then writes to `header.devicePath` directly, so a locally supplied or mispackaged archive can target files outside the supported resource and boot-resource locations.

Please restore an explicit allowlist here, including only narrowly defined newly supported paths, before trusting the header metadata.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -1055,133 +1037,69 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
-  private async prepareProtocolV2LocalMemoryHost({
-    features,
-    firmwareType,
-    availableInstallItems = this.buildProtocolV2InstallItems({
-      bootloaderBinary: this.prepareBootloaderBinary(),
-      fwBinaryMap: this.collectExplicitTargetBinaries(),
-    }),
+  private async runProtocolV2DirectArtifacts({
+    availableInstallItems,
+    announceDownload = true,
```
originalix, line 1033: P1: [Direct local V4 updates no longer reject missing requested firmware targets]

When `targetsToUpdate` includes both firmware and `resource`, this path filters only the binaries that were supplied and proceeds as soon as a resource archive exists. A call such as `targetsToUpdate: ['app_v1', 'resource']` with only `resourceArchiveBinary` therefore succeeds as a resource-only update.

Please call `getMissingProtocolV2FirmwareTargets()` (or an equivalent check) before transfer and return `FirmwareArtifactsNotPrepared` whenever a requested non-resource target has no local binary.

packages/core/src/index.ts
originalix, line 24: P2: [hd-core removes exported firmware-host helpers without compatibility handling]

This barrel no longer exports `prepareFirmwareUpdateV4MemoryHost` or the related memory-host types, and the implementation file is deleted in the same change. External updater hosts importing these published symbols from `@onekeyfe/hd-core` will fail to compile on upgrade.

Please keep a deprecated compatibility wrapper for one release, or document and version this as an intentional breaking change with a migration path.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1079,17 +1096,21 @@ export default class ReactNativeBleTransport {
           Log?.debug('monitor error ignored for stale transport: ', uuid, notifyTransactionId);
           return;
         }
+        if (
+          (this.getActiveProtocol(uuid) === 'V2' || this.acquiringProtocolV2.has(uuid)) &&
```
sidmorizon: # 🟠 P1 收口到 V2 的同时丢掉了 V1 原有的绑定失效分类

## 问题

本次提交把 monitor 里的陈旧绑定识别加了 `getActiveProtocol(uuid) === 'V2' || acquiringProtocolV2.has(uuid)` 前置条件，但 V1 分支（第 1126 行 `if (this.runPromise && this.runPromiseDeviceId === uuid)`）里并没有把 base 上原有的兜底补回来。base `onekey`@`3d5fcd3` 的同一段是：

```ts
let ERROR: BleDeviceBondError | BleCharacteristicNotifyError | BleTimeoutError =
  HardwareErrorCode.BleCharacteristicNotifyError;
if (error.reason?.includes('The connection has timed out unexpectedly')) ERROR = BleTimeoutError;
if (error.reason?.includes('Encryption is insufficient')) ERROR = BleDeviceBondError; // ← 已消失
```

这条 `Encryption is insufficient` 映射在 `df17363` 时由无条件的 `isNativeBleStaleBondError(error)` 早返回接管（走 `toBleStaleBondHardwareError` 同样得到 `BleDeviceBondError`），本次给早返回加了 V2 门槛后，V1 就彻底没有任何路径再产出 `BleDeviceBondError` 了。

同一处收口还影响 `remapError()`：base 无条件判定 `iosErrorCode === BleATTErrorCode.UnlikelyError`（即 CBError `14` peerRemovedPairingInformation）或 `reason === 'Peer removed pairing information'` 抛 `BlePeerRemovedPairingInformation`，现在只有 `mapProtocolV2StaleBond` 为真才生效，V1 acquire 连接失败会退化成 `BleConnectedError`。

## 影响

Classic / Mini / Touch / Pro 这些 Protocol V1 BLE 设备在系统里被解除配对后，notify 报 `Encryption is insufficient`，`ensureConnected` 收到的错误码从 `715 Bluetooth pairing failed` 变成了 `709 Characteristic Notify Error`。两者都在立即 reject 列表里，所以不会卡超时，但上层拿不到「配对失效」这个语义，无法再引导用户重新配对——这是本 PR 之前就存在、且与 Protocol V2 无关的 V1 行为被静默移除，违反 `AGENTS.md` 中「新增 Protocol V2 支持时必须保留 Protocol V1 行为」。iOS V1 上 CBError `14` 退化为 `BleConnectedError` 后还会进入 `isRetryableBleConnectionError` 的 6 次重试，最终以 `DeviceNotFound` 收场，同样丢失了原有语义。

本次新增的 `keeps native stale-bond write mapping out of Protocol V1 calls` / `keeps stale-bond subscribe mapping out of Protocol V1 acquire` 两个用例只锁住了「V1 不产出新错误码」，没有任何用例锁住「V1 仍然保留 base 的 `BleDeviceBondError` / `BlePeerRemovedPairingInformation`」。

## 建议

V2 门槛只用来控制**新增**的 fail-fast（`rememberStaleBondError` + `throwIfStaleBondError`），不要顺带删掉 V1 原有的错误分类：在 V1 分支恢复 `error.reason?.includes('Encryption is insufficient') → BleDeviceBondError`，并在 `remapError()` 中把 base 原有的 `iosErrorCode === 14` / `reason === 'Peer removed pairing information'` 判定保留为不受 `mapProtocolV2StaleBond` 约束的分支。补一条 V1 notify 报 `Encryption is insufficient` 时仍得到 `BleDeviceBondError` 的用例固化边界。

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2001,7 +1913,7 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         }
         lastError = new Error('Protocol V2 device is reachable but is not in bootloader mode');
       } catch (error) {
-        if (this.isProtocolV2ReconnectIdentityError(error)) {
+        if (isBleStaleBondHardwareError(error) || this.isProtocolV2ReconnectIdentityError(error)) {
```
sidmorizon: # 🟠 P1 重启轮询把「Ping 未响应」误当成陈旧绑定，会中断固件更新

## 问题

`isBleStaleBondHardwareError()` 的判定是 `errorCode === BleDeviceBondError || errorCode === BlePeerRemovedPairingInformation`（`packages/shared/src/HardwareError.ts:685`）。但 `BleDeviceBondError` 并不只来自原生加密/配对失败——两个 BLE transport 都会在「曾经确认过 V2 的 uuid 这次 Ping 没应答」时合成同一个码：

```ts
// packages/hd-transport-react-native/src/index.ts:2035
const isStaleV2Bond = expected === 'V2' && this.confirmedProtocolV2.has(uuid);
return ERRORS.TypedError(
  isStaleV2Bond ? HardwareErrorCode.BleDeviceBondError : HardwareErrorCode.RuntimeError,
  `Device protocol mismatch: expected ${expected}, but device did not respond ...`
);
```

`waitForProtocolV2BootloaderMode()` 的存在意义正是跨越设备重启窗口反复 `reconnectProtocolV2Device()` + 探测。而设备在这个窗口里恰好满足 `confirmedProtocolV2.has(uuid) === true` 且 Ping 不应答，于是 `detectProtocol()` 抛出的就是合成的 `BleDeviceBondError`，本行新增的判断会在**第一轮**就 `throw`。`waitForProtocolV2AppMode()`（第 2707 行）、`protocolV2SourceUpdateProcess()` 的重试（第 2198 行）以及 `isProtocolV2TerminalInstallStatusError()`（第 371 行）都引入了同一个谓词，问题一致。

## 影响

Pro2/Neo 走 BLE 做固件更新时，进入 bootloader 与装完回 App 这两个重启窗口本来是靠轮询兜住的；现在只要有一轮探测没应答，`firmwareUpdateV4` 就直接以 `BleDeviceBondError`（文案 `Bluetooth pairing failed`）失败退出，而不是继续等到 `PROTOCOL_V2_BOOTLOADER_RECONNECT_TIMEOUT`。用户在固件更新中途看到「蓝牙配对失败」，会被引导去系统里删除配对——设备此时可能正处于 bootloader/staging 中间态，这是固件写入路径上最不该做的事。本次新增的用例 `stops V4 bootloader reconnect polling on a stale BLE bond` 直接 mock `reconnectProtocolV2Device` 抛 `BlePeerRemovedPairingInformation` 并断言 `toHaveBeenCalledTimes(1)`，恰好把「不再轮询」这一行为固化了下来，却没有覆盖合成 `BleDeviceBondError` 的这条真实路径。

## 建议

这些位置需要的是「原生加密/配对确实失效」的强信号，而不是 transport 合成的探测失败码。建议在 V4 里改用只认 `BlePeerRemovedPairingInformation` 的窄谓词，或者让 transport 给真正来自 `isNativeBleStaleBondError()` 的错误带上可区分的标记（例如 `params.nativeStaleBond = true`），Core 侧按标记判断：

```ts
const isTerminalStaleBond = (error: unknown) =>
  (error as { errorCode?: unknown })?.errorCode ===
    HardwareErrorCode.BlePeerRemovedPairingInformation ||
  (error as { params?: { nativeStaleBond?: boolean } })?.params?.nativeStaleBond === true;
```

同时补一条「重启窗口内 `reconnectProtocolV2Device` 抛合成 `BleDeviceBondError` 时仍继续轮询」的用例。

[Claude session ↗](https://claude.ai/code/session_01TmjLbs6P7588HWGxEacfLo)

---

packages/hd-transport-react-native/src/index.ts
```diff
@@ -1079,17 +1096,21 @@ export default class ReactNativeBleTransport {
...
+          isNativeBleStaleBondError(error)
+        ) {
+          this.rememberStaleBondError(uuid, toBleStaleBondHardwareError(error));
+          return;
+        }
```
sidmorizon: **Classic/Touch/Pro Bluetooth devices no longer report the "pairing removed" error when their bond goes stale**

Severity: `non-severe`

The stale-pairing detection is applied only to Protocol V2 sessions (`getActiveProtocol(uuid) === 'V2' || acquiringProtocolV2.has(uuid)` at `packages/hd-transport-react-native/src/index.ts:1099-1105`), so a Protocol V1 device whose pairing was removed now reports a generic notification failure instead of the pairing error it used to report.
Impact: On iOS/Android, users of the older Bluetooth devices no longer get the "delete pairing information / re-pair" guidance; the connection just retries and eventually fails with a generic "device not found".

<details>
<summary>V1 branch of the notification handler lost its insufficient-encryption mapping</summary>

Before this PR, the Protocol V1 branch of `_monitorCharacteristic` mapped `error.reason?.includes('Encryption is insufficient')` to `HardwareErrorCode.BleDeviceBondError`, and `remapError` mapped `reason === 'Peer removed pairing information'` / `iosErrorCode === 14` to `HardwareErrorCode.BlePeerRemovedPairingInformation` for every protocol. Both mappings are now gated:

- the notification handler only converts native stale-bond errors when the active (or in-flight acquire) protocol is V2; the V1 `runPromise` branch at `packages/hd-transport-react-native/src/index.ts:1126-1135` now falls through to `BleCharacteristicNotifyError`.
- `remapError(error, mapProtocolV2StaleBond)` (`packages/hd-transport-react-native/src/index.ts:314-318`) only maps stale bonds when `expectedProtocol === 'V2'` or the endpoint previously confirmed V2, so V1 connects surface `BleConnectedError`.

`BleDeviceBondError` is in the terminal reject list in `packages/core/src/core/index.ts:1154`, while `BleCharacteristicNotifyError` is not, so the V1 flow now silently retries instead of surfacing the actionable error. AGENTS.md requires preserving Protocol V1 behavior when adding Protocol V2 support.

</details>

originalix (commented):
@wabicai

Reviewed commit `647b5bffc1dc`.

- **P1 · Protocol V2 resource packages can write outside supported device paths**
- **P1 · Direct resource updates can omit requested firmware targets**
- **P2 · Published firmware-host exports are removed without a compatibility period**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/890#issuecomment-5338994038)

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Not a silent firmware skip when both artifacts are supplied.

If `targetsToUpdate` includes `app_v1` *and* `applicationP1Binary` is actually passed, `runProtocolV2DirectArtifacts()` still collects that binary via `collectExplicitTargetBinaries()` / `filterProtocolV2LocalInstallItems()` and installs firmware together with the resource ZIP.

The reported case is only an incomplete request: `targetsToUpdate: ['app_v1', 'resource']` with *only* `resourceArchiveBinary`. The previous memory-host path threw `FirmwareArtifactsNotPrepared`; this path now filters to the binaries that exist and completes as a resource-only update.

App, CLI, firmware-updater-web, and the examples construct `targetsToUpdate` from the binaries they actually have, so this incomplete-payload shape does not occur on our hosts. No change in this PR.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Not a silent firmware skip when both artifacts are supplied.

If `targetsToUpdate` includes `app_v1` *and* `applicationP1Binary` is actually passed, `runProtocolV2DirectArtifacts()` still collects that binary via `collectExplicitTargetBinaries()` / `filterProtocolV2LocalInstallItems()` and installs firmware together with the resource ZIP.

The reported case is only an incomplete request: `targetsToUpdate: ['app_v1', 'resource']` with *only* `resourceArchiveBinary`. The previous memory-host path threw `FirmwareArtifactsNotPrepared`; this path now filters to the binaries that exist and completes as a resource-only update.

App, CLI, firmware-updater-web, and the examples construct `targetsToUpdate` from the binaries they actually have, so this incomplete-payload shape does not occur on our hosts. No change in this PR.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Not a silent firmware skip when both artifacts are supplied.

If `targetsToUpdate` includes `app_v1` *and* `applicationP1Binary` is actually passed, `runProtocolV2DirectArtifacts()` still collects that binary via `collectExplicitTargetBinaries()` / `filterProtocolV2LocalInstallItems()` and installs firmware together with the resource ZIP.

The reported case is only an incomplete request: `targetsToUpdate: ['app_v1', 'resource']` with *only* `resourceArchiveBinary`. The previous memory-host path threw `FirmwareArtifactsNotPrepared`; this path now filters to the binaries that exist and completes as a resource-only update.

App, CLI, firmware-updater-web, and the examples construct `targetsToUpdate` from the binaries they actually have, so this incomplete-payload shape does not occur on our hosts. No change in this PR.

packages/core/src/index.ts
wabicai: No compatibility shim needed.

App, CLI, firmware-updater-web, and the examples were migrated in lockstep away from `prepareFirmwareUpdateV4MemoryHost`. Nothing we ship still imports that public helper from `@onekeyfe/hd-core`. We will not keep a deprecated wrapper for this PR.

packages/core/src/protocols/protocol-v2/resources.ts
wabicai: Intentional, not a regression for Classic/Touch/Pro.

Protocol V2 resource writes use the signed RESC header `devicePath` (`flexible_metadata`) as the device contract. Host-side path allowlisting was removed so new firmware destinations do not require an SDK allowlist bump. Traversal, empty segments, and control characters still fail in `isSafeResourceDevicePath()` before the first `FilesystemFileWrite`.

This does not affect Protocol V1 devices.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Not blocking for this PR.

Live on-device compare is only defined for `boot_resource.okpkg` after early boot consumes the staging file. Other `.okpkg.staging` packages do not have a documented live path we can strip `.staging` against. We will not generalize that compare path without a firmware contract.

packages/core/src/device/DeviceCommands.ts
wabicai: Out of scope for this PR.

Stale BLE bond / pairing-error handling was moved to https://github.com/OneKeyHQ/hardware-js-sdk/pull/893. The current #890 tip no longer contains those transport/Core changes.

packages/hd-transport-react-native/src/index.ts
wabicai: Out of scope for this PR.

Stale BLE bond / pairing-error handling was moved to https://github.com/OneKeyHQ/hardware-js-sdk/pull/893. The current #890 tip no longer contains those transport/Core changes.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Out of scope for this PR.

Stale BLE bond / pairing-error handling was moved to https://github.com/OneKeyHQ/hardware-js-sdk/pull/893. The current #890 tip no longer contains those transport/Core changes.

packages/hd-transport-react-native/src/index.ts
wabicai: Out of scope for this PR.

Stale BLE bond / pairing-error handling was moved to https://github.com/OneKeyHQ/hardware-js-sdk/pull/893. The current #890 tip no longer contains those transport/Core changes.

originalix (commented):
@wabicai

Reviewed commit `302f9ff916b3`.

- **P1 · Protocol V2 resource packages can write outside supported device paths**
- **P1 · Direct resource updates can omit requested firmware targets**
- **P1 · Plan-driven example updates drop device identity and version safeguards**
- **P2 · Published firmware-host exports are removed without a compatibility period**

[View the full review report](https://github.com/OneKeyHQ/hardware-js-sdk/pull/890#issuecomment-5338994038)

originalix (approved):

PR #891, opened 2026-08-19
fix: report desktop USB and BLE device disconnects (OK-60486)

originalix:
## Problem

`DEVICE.DISCONNECT` never fired on desktop, for either transport. Consumers that
track connection state therefore kept treating a removed device as connected —
see OK-60486, where the wallet-list connected dot stayed lit after unplugging,
but only once the device had been interacted with.

Three delivery paths exist in hd-core and all three were dead on desktop:

- `onTransportDeviceDisconnectHandler` ← `TRANSPORT_EVENT.DEVICE_DISCONNECT`
- `DevicePool.reportDeviceChange` ← `DeviceConnector.listen()`, which is only
  ever re-entered by itself and has no external caller
- `DevicePool._sendDisconnectMessage`, fed only by the above

## Fixes

**WebUSB — no path existed at all.** `TransportManager.configure` initialised the
webusb branch as `transport.init(WebUsbLogger)` with no emitter, unlike the
`react-native`, `node-usb` and `desktop-web-ble` branches, and `webusb.ts` had
zero `TRANSPORT_EVENT` references. Its `navigator.usb` disconnect listener only
marked the protocol cache stale. Now the emitter is passed and that listener also
emits `TRANSPORT_EVENT.DEVICE_DISCONNECT`.

**Electron BLE — the path was disarmed between calls.** The only emit site was an
IPC subscription registered inside `acquire()`, and the end-of-call
`releaseLogical()` → `cleanupDeviceState()` disposed it, so a drop during the
keep-alive window went unobserved. `acquire()` also overwrote `disconnectCleanups`
without disposing the previous handler, leaking one live listener per acquire —
which is what the `connectionToken` guard existed to paper over. Both are replaced
by a single transport-lifetime subscription created in `init()`.

**Keep-alive releases are reported too.** The main process reclaims idle links on
its own timer and broadcast the same event as a real drop. `BLE_DEVICE_DISCONNECTED`
now carries a `reason`, and every link drop is reported: consumers track whether a
BLE link is live, and a link we closed ourselves is still a closed link. Nothing
reconnects on its own — background portfolio sync only piggybacks on an already-open
link — so the state settles once after the last interaction rather than oscillating.
`reason` is retained for diagnostics; behaviour is uniform.

## Tests

Replaced `ignores a delayed disconnect event from the previous BLE connection`,
which encoded the leaked-handler design, with four cases covering the new contract:

- host subscription is created once for the transport lifetime, not per acquire
- a device dropping **after a logical release** is reported (the OK-60486 regression;
  fails on the old code, where `cleanupDeviceState` had already deleted the
  `notificationTokens` entry the token guard compared against)
- an idle keep-alive release is reported, and cached link state is dropped
- a disconnect with no `reason` is treated as a real drop, so an older host bridge
  that omits the field behaves as before

Full suite green across 18 packages.

## Verified on hardware

macOS desktop, OneKey Classic 1S over both transports: USB unplug, real BLE drop,
and BLE keep-alive expiry all clear the indicator. Confirmed in the app log via
`device disconnected, untracked` with a non-zero removed-key count, where the same
build before the fix logged the keep-alive timeout and nothing after it.

## Release

`1.2.0-alpha.155`.


sidmorizon:
@codex review

sidmorizon:
@codex security review

wabicai (approved):

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -231,9 +234,48 @@ export default class ElectronBleTransport {
...
   }
 
+  /**
+   * One host subscription for the whole transport lifetime. init() can run
+   * again after an SDK reset, so drop the previous listener first rather than
+   * stacking duplicates.
+   */
+  private subscribeHostDisconnects() {
+    this.hostDisconnectCleanup?.();
+    this.hostDisconnectCleanup = window.desktopApi?.nobleBle?.onDeviceDisconnected(
```
sidmorizon, line 249: **Bluetooth disconnect notifications pile up and fire repeatedly after the SDK is restarted**

Severity: `non-severe`

A new Bluetooth disconnect listener is registered every time the SDK is started (`subscribeHostDisconnects()` at `packages/hd-transport-web-device/src/electron-ble-transport.ts:237`) while the one from the previous start is never removed, so one unplug event is reported several times.

Impact: After an SDK restart, a single Bluetooth device drop produces duplicate disconnect notifications to the app, and each restart permanently retains another listener plus the old transport object it belongs to.

<details>
<summary>Why the "drop the previous listener first" guard does not help across SDK restarts</summary>

`subscribeHostDisconnects()` disposes `this.hostDisconnectCleanup` before subscribing, which only protects against `init()` running twice on the *same* instance. In practice `TransportManager.setTransport` constructs a brand-new transport object on every SDK `init()` (`packages/core/src/data-manager/TransportManager.ts:123-139`, called from `packages/core/src/core/index.ts:1775-1786`), and `TransportManager.configure()` then calls `init()` on that new object. The previous instance keeps its own `hostDisconnectCleanup`, which nothing ever invokes — `ElectronBleTransport` has no `stop()`/`dispose()` method, so the disposal path in `Core.disposeResources` (`TransportManager.getTransport()?.stop?.()`) is a no-op for it.

Because `this.emitter` on every instance is the same static `DevicePool.emitter`, each stale listener still emits `TRANSPORT_EVENT.DEVICE_DISCONNECT`, so `onTransportDeviceDisconnectHandler` (`packages/core/src/core/index.ts:1367-1379`) runs once per leaked instance and re-emits `DEVICE.DISCONNECT` to the app each time. The previous per-acquire subscription design disposed its listeners in `cleanupDeviceState`, so this cross-instance accumulation is new. Note `webusb.ts` deliberately avoids the same problem by routing the module-scoped listener to `activeWebUsbTransport` only.

</details>

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -231,9 +234,48 @@ export default class ElectronBleTransport {
...
+        // and a link we closed ourselves is still a closed link. Nothing
+        // reconnects on its own, so this settles once until the user acts.
+        // `reason` is carried for diagnostics only — behaviour is uniform.
+        this.Log?.debug(
+          '[Electron BLE] Device link dropped:',
+          uuid,
+          disconnectedDevice.reason ?? EBleDisconnectReason.DeviceDisconnected
+        );
+
+        this.emitter?.emit(TRANSPORT_EVENT.DEVICE_DISCONNECT, {
```
sidmorizon, line 270: # 🟠 P1 空闲保活回收被上报成设备真实掉线

## 问题

订阅现在是 transport 生命周期级的，`release()` 不再注销它，同时这里丢弃了 `disconnectedDevice.reason`，对 `IdleKeepAlive` 和 `DeviceDisconnected` 发出完全相同的 `TRANSPORT_EVENT.DEVICE_DISCONNECT`。

主进程的逻辑 release 会 `armIdleDisconnect(deviceId)`（`noble-ble-handler.ts`，`BLE_IDLE_DISCONNECT_MS = 3 * 60_000`），到点主动断链并广播 `IdleKeepAlive`。改动前 `releaseLogical()` → `cleanupDeviceState()` 已经注销了 per-acquire 的订阅，这个广播在 renderer 侧无人接收；改动后它会一路上报。

## 影响

于是每次正常用完设备并 release 之后约 `3` 分钟，即使设备开着、就在手边、用户什么都没做，`core/index.ts` 的 `onTransportDeviceDisconnectHandler` 都会命中缓存里的 Device，执行 `device.markTransportDisconnected()`（`deviceAcquired = false` + `invalidateProtocolV2RuntimeState()`，Pro2 的 `protocolV2RuntimeContext` 与 pre-initialized 状态被清空），并向 App 抛出 `DEVICE.DISCONNECT`。

这是常态路径而非边界情况：桌面端 BLE 每一次设备交互后都会触发一次，UI 会把在线设备标记为已断开，Protocol V2 运行时状态也要重新建立。

## 建议

`reason` 已经一路传到这里，让它参与决策而不是只进 log：要么对 `IdleKeepAlive` 不发 `DEVICE_DISCONNECT`（只清理本地链路状态，下次 acquire 自然重连），要么把 `reason` 透传给上层由 core/App 区分。

```ts
const reason = disconnectedDevice.reason ?? EBleDisconnectReason.DeviceDisconnected;
this.cleanupDeviceState(uuid);
if (reason === EBleDisconnectReason.IdleKeepAlive) return;
this.emitter?.emit(TRANSPORT_EVENT.DEVICE_DISCONNECT, { ... });
```

[Claude session ↗](https://claude.ai/code/session_015NoiUSeJMRQwpJin3SaQYd)

---

packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -394,9 +401,14 @@ function armIdleDisconnect(
       const pending = disconnectDevice(deviceId)
         .then(() => {
           // A call is still in flight here; it must reject, not hang.
+          // Both 'idle' and 'busy-backstop' report IdleKeepAlive on purpose:
+          // we reclaimed the link ourselves and have no evidence the device
+          // left, so consumers must not mark it disconnected. The in-flight
```
sidmorizon, line 406: # 🟡 P2 新增注释与本 PR 最终行为相反

## 问题

这段注释断言 `consumers must not mark it disconnected`，`packages/shared/src/constants.ts` 里 `EBleDisconnectReason` 的 doc 也写着 `consumers must not treat it as "the device is gone"`。但同一个 PR 的最后一个 commit（`922cb4e`）里，唯一的消费者 `electron-ble-transport.ts` 的 `subscribeHostDisconnects()` 恰恰对 `IdleKeepAlive` 发出了和真实掉线完全一致的 `TRANSPORT_EVENT.DEVICE_DISCONNECT`，注释里写的 `reason` 契约在代码中并不成立。

同一处还说 in-flight call `still fails, via the renderer dropping its device state`——这句描述的其实正是新消费端 `cleanupDeviceState()` 的行为，与前一句「不要当作断开」互相矛盾。

## 影响

`EBleDisconnectReason` 是本 PR 新引入的跨进程契约，这两处注释是它唯一的规范说明。后续维护者按注释去实现新的 renderer 或修改现有分支时，会得到与线上行为相反的预期；调试「设备明明在线却报断开」时也会因为注释而排除掉这条真实路径。

## 建议

按最终采纳的语义统一表述。若保留「所有 link drop 一律上报」，把两处注释改成说明 `reason` 仅用于诊断/日志、不影响行为；若采纳按 `reason` 分流的做法，则让消费端与注释一致。二选一即可，但不要让两者继续并存。

[Claude session ↗](https://claude.ai/code/session_015NoiUSeJMRQwpJin3SaQYd)

---

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -231,9 +234,48 @@ export default class ElectronBleTransport {
...
+    this.hostDisconnectCleanup = window.desktopApi?.nobleBle?.onDeviceDisconnected(
+      (disconnectedDevice: { id: string; name: string; reason?: EBleDisconnectReason }) => {
+        const uuid = disconnectedDevice?.id;
+        if (!uuid) return;
+
+        // The link is gone, so renderer-side state must go with it; the next
+        // acquire reconnects from scratch.
+        this.cleanupDeviceState(uuid);
+
+        // Every link drop is reported, including the main process reclaiming
```
originalix, line 258: P1: [Idle keep-alive BLE releases are reported as real device removals]

When the main process frees an idle keep-alive link it now sends `reason: IdleKeepAlive` specifically to say "the transport link was reclaimed, but the peripheral did not necessarily leave".

This handler ignores that distinction and always emits `TRANSPORT_EVENT.DEVICE_DISCONNECT`, which core converts into `DEVICE.DISCONNECT`. That means a normal logical release followed by the idle timer will look the same as an unplug/power-off event: cached runtime state is cleared and consumers are told the device disappeared even though it may still be present.

Please branch on `disconnectedDevice.reason` here. It is fine to clean local link state for every drop, but only real `DeviceDisconnected` cases should be surfaced as the device-removal event, or this needs a separate link-closed signal for the idle path.

packages/hd-transport-web-device/src/electron-ble-transport.ts
```diff
@@ -231,9 +234,48 @@ export default class ElectronBleTransport {
...
+
     this.Log?.debug('[Electron BLE] Transport initialized');
   }
 
+  /**
+   * One host subscription for the whole transport lifetime. init() can run
+   * again after an SDK reset, so drop the previous listener first rather than
+   * stacking duplicates.
+   */
+  private subscribeHostDisconnects() {
```
originalix, line 247: P2: [The new host-level BLE disconnect listener is never torn down]

`subscribeHostDisconnects()` turns this into a transport-lifetime IPC subscription, but there is still no matching `stop()`/dispose path that invokes `hostDisconnectCleanup`.

The trigger is any SDK dispose/re-init or transport switch: core already calls `transport.stop?.()` during teardown, but this class does not implement it, so each new instance leaves another `onDeviceDisconnected` listener registered in the renderer. The wrong mechanism is the one-way setup with no symmetric unsubscribe, and the impact is stale transport instances being retained and future disconnects being handled more than once.

Please add a `stop()` implementation that calls `hostDisconnectCleanup`, clears the emitter reference, and resets any transport-lifetime listener state so teardown really releases this subscription.

originalix:

@originalix

Reviewed commit `922cb4e29b2d`. This report always reflects the latest reviewed changes.

### Review summary

The PR rewires disconnect propagation for Electron BLE and WebUSB so transport-layer link drops reach core, adds a BLE disconnect-reason enum to distinguish idle keep-alive releases from real peripheral drops, and bumps internal package versions for the next alpha publish.

**What needs attention:** Verify that desktop BLE idle keep-alive releases do not masquerade as physical unplug events, and ensure the new transport-lifetime disconnect subscription is torn down during SDK dispose/re-init.

### Issues to address

- **P1 · Idle keep-alive BLE releases are reported as real device removals**
  - **Impact:** The main-process BLE handler now sends `reason: IdleKeepAlive` specifically to distinguish an internal idle-link reclaim from a peripheral actually dropping. The renderer transport ignores that distinction and always emits `TRANSPORT_EVENT.DEVICE_DISCONNECT`, which core turns into `DEVICE.DISCONNECT`. On the normal keep-alive timeout path, consumers will be told the device disappeared, cached runtime state is cleared, and UI/device state will flap even though the peripheral is still present.
  - **Suggested action:** Branch on `disconnectedDevice.reason` in the renderer transport: keep local link cleanup for every drop, but only emit the device-removal event for `DeviceDisconnected` (or send a separate link-closed signal if the UI still needs telemetry for idle releases).
- **P2 · The new host-level BLE disconnect listener is never torn down**
  - **Impact:** `subscribeHostDisconnects()` registers a transport-lifetime `window.desktopApi.nobleBle.onDeviceDisconnected(...)` callback, but the transport still has no `stop()` implementation. Core already calls `transport.stop?.()` during dispose, so repeated SDK init/dispose cycles or transport switches will leave stale IPC listeners behind. That retains dead transport instances and can duplicate future disconnect handling.
  - **Suggested action:** Add a `stop()` implementation that invokes `hostDisconnectCleanup`, clears the emitter reference, and resets any transport-lifetime listener state so SDK disposal fully releases the renderer-side BLE subscription.

### Validation gaps

- I did not complete automated test execution. A targeted `yarn jest packages/hd-transport-web-device/__tests__/electron-ble-transport.test.ts --runInBand` attempt failed because the `jest` command is not available in this workspace.
- The PR adds WebUSB disconnect propagation but only extends Electron BLE tests; there is no automated coverage here for the new `navigator.usb` disconnect path.

PR #892, opened 2026-08-20
fix: scope stale BLE bond handling to protocol v2

originalix:
## Summary

- Scope the new stale-bond fast-fail mapping to strict or previously confirmed Protocol V2 connections in React Native BLE and Electron BLE.
- Restore Protocol V1 write, probe, subscribe, and error propagation behavior so Classic, Mini, Touch, and Pro do not enter the new recovery path.
- Keep ATT 0x0E Unlikely Error outside stale-bond classification; retain ATT 5/15 and iOS peer-removed signals for V2.
- Preserve the distinct peer-removed error through Core only for Protocol V2 connection handling.
- Make FirmwareUpdateV4 fail fast on stale-bond errors during resource checks, reconnect polling, file transfer, install polling, and final verification; FirmwareUpdateV3 keeps its previous wrapping behavior.
- Bump all aligned SDK packages from 1.2.0-alpha.160 to 1.2.0-alpha.161.

## Scope

Protocol V2 is the actively probed transport boundary used by Pro 2 and Neo. Device names and PIDs are not used to infer this behavior. Firmware handling is limited to FirmwareUpdateV4; V1/V3 behavior is covered by negative regression tests.

This PR targets fix/pro2-ble-name-space, the reviewed branch of #890, so the diff shows only the follow-up hardening and alpha version bump.

No package has been published. Publication will wait for diff approval.

## Validation

- Core focused tests: 396 passed
- React Native BLE focused tests: 55 passed
- Electron BLE focused tests: 24 passed
- Focused ESLint: passed
- yarn agent:check --profile commit: passed
- yarn agent:check --profile pr: passed


packages/core/src/api/FirmwareUpdateV3.ts
```diff
@@ -566,7 +566,6 @@ export default class FirmwareUpdateV3 extends FirmwareUpdateBaseMethod<FirmwareU
           HardwareErrorCode.BlePermissionError,
           HardwareErrorCode.BleLocationError,
           HardwareErrorCode.BleDeviceBondError,
-          HardwareErrorCode.BlePeerRemovedPairingInformation,
```
originalix, line 569: 你为什么要直接对它做删除？

packages/core/src/core/index.ts
```diff
@@ -950,6 +950,14 @@ export function isMissingDetectedProtocolV2Error(method: BaseMethod, error: unkn
   );
 }
 
+export function isProtocolV2PeerRemovedPairingError(method: BaseMethod, error: unknown) {
```
originalix, line 953: P1: [Known Protocol V2 stale bonds still fall through to polling when connectProtocol is omitted]

This helper only treats `BlePeerRemovedPairingInformation` as terminal when the raw method payload says `connectProtocol === 'V2'`.

That misses the common BLE path where the payload omits `connectProtocol`, but `Device.acquire()` still uses the cached descriptor protocol as the strict protocol. In that case a known Pro2/Neo can correctly return `BlePeerRemovedPairingInformation`, yet `ensureConnected()` will not reject it here and keeps polling until a generic timeout or not-found error hides the real re-pair failure.

Please key this check off the resolved device protocol, or at least the descriptor protocol that is actually being passed into `device.acquire()`, instead of only the raw payload field.

originalix:

@originalix

Reviewed commit `357720044f18`. This report always reflects the latest reviewed changes.

### Review summary

This PR narrows BLE stale-bond handling to Protocol V2-specific paths across React Native BLE, Electron BLE, core connection orchestration, and Protocol V2 firmware-update recovery, then bumps internal package versions to 1.2.0-alpha.161. The intended design is sound: transport adapters classify native BLE failures, core decides retry versus terminal behavior, and firmware-update polling should stop immediately on explicit stale-bond errors.

**What needs attention:** Verify stale-bond behavior when callers rely on cached Protocol V2 state, force active protocol detection, or reconnect during Protocol V2 firmware update flows on both React Native and Electron BLE.

### Issues to address

- **P1 · Known Protocol V2 stale bonds still fall through to polling when connectProtocol is omitted**
  - **Impact:** Core now treats `BlePeerRemovedPairingInformation` as terminal only when `method.payload.connectProtocol === 'V2'`. That misses the common BLE path where the payload omits `connectProtocol`, but `Device.acquire()` still uses the cached descriptor protocol (`originalDescriptor.protocolType`) as the strict protocol. In that case a known Pro2/Neo can correctly raise `BlePeerRemovedPairingInformation`, yet `ensureConnected()` will not reject it immediately and instead keeps polling until a generic timeout or not-found error replaces the real re-pair signal.
  - **Suggested action:** Derive the terminal Protocol V2 stale-bond check from the resolved device protocol or the descriptor protocol actually passed into `device.acquire()`, not only from the raw method payload.
- **P2 · Forced re-detection and protocol-hinted V2 acquires stop classifying stale bonds**
  - **Impact:** Both BLE transports gate stale-bond remapping on `expectedProtocol === 'V2'` or an in-memory `confirmedProtocolV2` cache. When callers use `forceProtocolDetection` or only pass `protocolHint: 'V2'`, a fresh transport instance reaches connect/subscribe with that gate turned off. A real Pro2/Neo stale bond then degrades into a generic setup error instead of the explicit re-pair errors this PR is meant to preserve for Protocol V2.
  - **Suggested action:** Include the V2 hint or the caller's cached device protocol in the stale-bond gating for the acquire/setup phase, while still keeping generic V1 setup failures unmapped.

### Validation gaps

- I ran `npx jest --runInBand packages/core/__tests__/DeviceCommands.test.ts packages/core/__tests__/device-lifecycle-events.test.ts packages/core/__tests__/protocol-v2.test.ts packages/hd-transport-react-native/src/__tests__/bleStaleBond.test.ts packages/hd-transport-react-native/src/__tests__/protocolV2Link.test.ts packages/hd-transport-web-device/__tests__/electron-ble-transport.test.ts`.
- That run did not provide a clean signal because the workspace currently resolves some package-name imports against stale `dist` outputs, which surfaces unrelated runtime import failures such as missing `canonicalizePro2BleAdvertisementName`, `isBleStaleBondHardwareError`, and `EBleDisconnectReason` exports. I did not tie those failures to this PR diff, so the review conclusions above rely on source inspection of the changed paths.

wabicai (approved):

PR #893, opened 2026-08-20
fix: fail fast on Pro2/Neo stale BLE bonds

wabicai:
## Summary

Split stale BLE pairing/bond handling out of #890. After a Pro2/Neo wipe or reflash, iOS can keep an old OS bond. Transport now fails immediately on native ATT encryption / peer-removed pairing errors instead of waiting through Protocol V2 probe timeouts. Core stops polling on 719 for confirmed V2 devices even when the caller omits `connectProtocol`.

## Intent and context

#890 is BLE name canonicalization and Protocol V2 firmware/resource updates. Bond-error work landed on that branch and should ship separately.

## Root cause

GATT connect still succeeds with a stale iOS bond. The failure is on CCCD/write (`Encryption is insufficient` / `Peer removed pairing information`). The previous path swallowed that during protocol detection, then Core kept polling when `connectProtocol` was omitted.

## Design decisions

- Map native stale-bond signals in React Native and Electron BLE transports.
- Scope 719 as a terminal Core error only when the resolved protocol is V2: `payload.connectProtocol ?? device.originalDescriptor.protocolType`.
- Leave 715 on the existing generic terminal list.
- Do not treat generic ATT UnlikelyError (0x0E) as a stale bond.

## Compatibility and risk

Additive error mapping. First-time unpaired discovery without a confirmed V2 protocol still does not treat 719 as terminal. Classic/Touch/Pro V1 explicit connects are unchanged.

## Hardware coverage

Unit coverage for native mapping, Protocol V2 acquire fail-fast, Core 719 terminal rules including cached V2 without `connectProtocol`. Physical Pro2/Neo reset-then-reconnect on iOS still needs a device check.

## Test plan

- [x] `device-lifecycle-events` including cached V2 + omitted `connectProtocol`
- [ ] Physical Pro2/Neo: wipe/reset while iOS still bonded, confirm re-pair prompt instead of timeout/`DeviceNotFound`

### DIFF
--- a/packages/core/src/api/FirmwareUpdateV4.ts
+++ b/packages/core/src/api/FirmwareUpdateV4.ts
@@ -1,4 +1,11 @@
-import { EDeviceType, ERRORS, HardwareError, HardwareErrorCode, wait } from '@onekeyfe/hd-shared';
+import {
+  EDeviceType,
+  ERRORS,
+  HardwareError,
+  HardwareErrorCode,
+  isBleStaleBondHardwareError,
+  wait,
+} from '@onekeyfe/hd-shared';
 import JSZip from 'jszip';
 import {
   DeviceRebootType,
@@ -372,11 +379,12 @@ const isProtocolV2FirmwareStatusEndpointUnavailable = (error: unknown) => {
 };
 
 const isProtocolV2TerminalInstallStatusError = (error: unknown) =>
-  error instanceof HardwareError &&
-  (error.errorCode === HardwareErrorCode.ActionCancelled ||
-    error.errorCode === HardwareErrorCode.FirmwareError ||
-    error.errorCode === HardwareErrorCode.FirmwareVerificationFailed ||
-    error.params?.firmwareUpdateCode === PROTOCOL_V2_INSTALL_STATUS_CONFLICT_CODE);
+  isBleStaleBondHardwareError(error) ||
+  (error instanceof HardwareError &&
+    (error.errorCode === HardwareErrorCode.ActionCancelled ||
+      error.errorCode === HardwareErrorCode.FirmwareError ||
+      error.errorCode === HardwareErrorCode.FirmwareVerificationFailed ||
+      error.params?.firmwareUpdateCode === PROTOCOL_V2_INSTALL_STATUS_CONFLICT_CODE));
 
 const isProtocolV2TargetStatusFinished = (status: ProtocolV2FirmwareUpdateStatusTarget['status']) =>
   normalizeProtocolV2TargetStatus(status) === PROTOCOL_V2_TARGET_STATUS_FINISHED;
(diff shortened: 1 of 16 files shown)


packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -372,11 +379,12 @@ const isProtocolV2FirmwareStatusEndpointUnavailable = (error: unknown) => {
 };
 
 const isProtocolV2TerminalInstallStatusError = (error: unknown) =>
-  error instanceof HardwareError &&
-  (error.errorCode === HardwareErrorCode.ActionCancelled ||
-    error.errorCode === HardwareErrorCode.FirmwareError ||
-    error.errorCode === HardwareErrorCode.FirmwareVerificationFailed ||
-    error.params?.firmwareUpdateCode === PROTOCOL_V2_INSTALL_STATUS_CONFLICT_CODE);
+  isBleStaleBondHardwareError(error) ||
```
originalix, line 382: P1: [Firmware update now aborts on normal confirmed-V2 probe misses]

During a Pro2/Neo reboot, `device.acquire('V2')` can legitimately fail with `BleDeviceBondError` even when pairing is still valid, because both BLE transports already use that code for the "previously confirmed V2 device did not answer Ping yet" path.

Routing every `isBleStaleBondHardwareError()` result through the firmware-update terminal branches makes those normal reboot windows look like an unrecoverable stale bond, so bootloader polling / status polling / final reconnect stop retrying and the update fails early.

Please narrow this fast-fail predicate to the explicit stale-bond signal (`BlePeerRemovedPairingInformation` or another transport-native stale-bond marker), and keep the generic `BleDeviceBondError` path retryable here.

originalix:

@wabicai

Reviewed commit `40720c1b43ce`. This report always reflects the latest reviewed changes.

### Review summary

This PR adds explicit stale-BLE-bond detection for Protocol V2 in the React Native and Electron BLE transports, propagates a dedicated peer-removed pairing error into core connection handling, and updates Protocol V2 firmware-update flows to stop retrying when BLE pairing is believed to be invalid.

**What needs attention:** Narrow the firmware-update fast-fail path so only explicit stale-bond signals stop reconnect polling; ordinary confirmed-V2 probe misses still need to retry through reboot windows.

### Issues to address

- **P1 · Firmware update now aborts on normal confirmed-V2 probe misses**
  - **Impact:** `BleDeviceBondError` is not a pure stale-bond signal in this codebase: both BLE transports already use it when a previously confirmed Protocol V2 device simply does not answer `Ping` yet. By treating every `isBleStaleBondHardwareError()` result as terminal inside `FirmwareUpdateV4`, the PR turns ordinary reboot windows during bootloader/app transitions into unrecoverable update failures.
  - **Suggested action:** Use a narrower terminal predicate in `FirmwareUpdateV4` that only fast-fails on the explicit peer-removed pairing signal or another transport-native stale-bond marker, while leaving generic `BleDeviceBondError` reconnect misses retryable.

### Validation gaps

- I did not run device-side or simulator-based firmware-update flows in this review.
- The added tests cover the new explicit stale-bond path, but they do not cover the pre-existing `BleDeviceBondError` path that represents a temporary confirmed-V2 probe miss during reconnect.

PR #896, opened 2026-08-20
v1.2.0-alpha.176 fix: harden Protocol V2 firmware install polling

wabicai:
## Summary

- keep Protocol V2 firmware confirmation, data transfer, installation, and completion states aligned with device responses and current install status
- bound queued Protocol V2 calls by their original timeout so stale status requests are never sent after the active call settles
- refresh cached and uncached React Native BLE GATT connections before firmware install polling, including the bootloader reboot handoff
- release all Hardware SDK packages as `1.2.0-alpha.176`

## Test coverage

- added focused Core regressions for USB/BLE Request writes, current-install status, stale status, reconnect, App-mode verification, cancellation, and terminal responses
- added shared transport coverage for queued-call timeout behavior
- added React Native BLE coverage for cached and uncached install reconnects without Protocol V2 Ping

## Verification

- `yarn agent:check --profile pr`
- real iOS device build with `ENABLE_NATIVE_BACKGROUND_THREAD=true`
- Pro2 BLE firmware update including bootloader handoff reached completion on device

## Pre-landing review

No blocking findings. No public API, protobuf, dependency, or generated-output changes.


packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2978,55 +3000,31 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
...
     try {
-      await commands.typedCall('DeviceFirmwareUpdateRequest', 'Success', {});
-      this.protocolV2InstallRequestConfirmed = true;
-      this.postProgressMessage(1, 'installingFirmware');
-      Log.log('[FirmwareUpdateV4] firmware install confirmed by device response');
+      await commands.call(
+        'DeviceFirmwareUpdateRequest',
+        {},
+        {
+          returnAfterWrite: true,
```
originalix, line 3008: P1: [Delayed firmware request failures are swallowed]

`DeviceFirmwareUpdateRequest` now returns after the write and only records a delayed `Success`. If the device responds with `Failure`, the Protocol V2 session clears the pending write-only response as a common terminal message without calling this Success-only callback. The next status poll can then consume that Failure as its own response, treat it as an empty status list, and continue until timeout instead of reporting the install error.

Propagate delayed non-Success terminal responses into the firmware-update flow and abort polling with the original typed error. Please add a regression test for a `Failure` response arriving during the first status poll.

originalix:

@wabicai

Reviewed commit `a5152148317e`. This report always reflects the latest reviewed changes.

### Review summary

The PR changes Protocol V2 firmware installation to return immediately after writing the install request, deferring terminal response handling into status polling. It also adds total timeout accounting for queued Protocol V2 calls, refreshes React Native BLE connections during install recovery, expands install-polling tests, and synchronizes package versions.

**What needs attention:** Verify that delayed firmware-install responses preserve the previous error and interaction semantics, especially when the device returns Failure instead of Success.

### Issues to address

- **P1 · Delayed firmware request failures are swallowed**
  - **Impact:** DeviceFirmwareUpdateRequest now uses returnAfterWrite with a delayed Success callback. When the device later returns Failure, ProtocolV2Session treats Failure as a common terminal response, clears the pending write-only response, and does not invoke the Success-only callback. The next status poll can therefore receive the Failure as if it were its own response, interpret it as an empty status list, and continue polling instead of reporting the original install error.
  - **Suggested action:** Propagate any delayed non-Success terminal response to the firmware-update flow and abort polling with the same typed error, or keep a response-waiting call for paths that must surface request failures. Add a regression test for a Failure response arriving during the first status poll.

### Validation gaps

- The new tests cover successful write-first polling, reconnects, cancellation, and queue timeout behavior, but do not cover a delayed `Failure` response from `DeviceFirmwareUpdateRequest`.
- Physical-device validation is still needed for WebUSB and BLE cases where the loader disconnects before or after the request response.
- No new outbound network access was introduced.

originalix (approved):

packages/core/src/api/FirmwareUpdateV4.ts
originalix: 复核后确认这个问题确实存在，但影响边界是：部分延迟返回的 `Failure` 会丢失原始错误并最终表现为约 5 分钟的 `FirmwareInstallTimeout`，不会造成升级假成功、刷错固件或设备损坏，因此按 P2 已知问题跟进，不作为当前合入阻塞项。

这版先合入观察 Pro 2 / Neo 在 USB、WebUSB 和 BLE 下的实际升级效果，后续单独修复延迟终端响应的错误透传，并补对应的 `Failure` 回归测试。该 thread 暂时保留，方便后续追踪。

PR #899, opened 2026-08-25
v1.2.1 fix: harden Protocol V2 BLE install completion

wabicai:
## Summary

- Follow up on #897 by generalizing Protocol V2 BLE install recovery from coprocessor-only handling to bootloader, P1, P2, coprocessor, and SE01-SE04 targets.
- Treat a real BLE disconnect during the install Request, status polling, or reconnect as install evidence only after the device reconnects successfully.
- Require four identical, complete `FINISHED` snapshots before accepting a fast BLE install when `IN_PROGRESS` was hidden by the disconnect.
- Keep command response timeouts out of the disconnect evidence path, while correctly recognizing React Native BLE link timeouts as physical disconnects.
- Align all published packages and internal package references to stable version `1.2.1`.

## Behavior and safety

- A response timeout alone cannot complete an installation.
- A real disconnect alone cannot complete an installation.
- Completion requires a successful reconnect plus four stable, complete target snapshots.
- The fallback intentionally supports same-version reinstallations, where version comparison cannot prove completion. Without a firmware-provided install generation or nonce, an unrelated physical BLE disconnect followed by four stale snapshots is a residual theoretical ambiguity.

## Test coverage

- Target matrix: bootloader, P1, P2, coprocessor, and SE01-SE04.
- Disconnect evidence: install Request, status polling, and reconnect phases.
- Negative paths: Request/status/reconnect response timeouts, cancellation, Stage rejection, and stale `FINISHED` status after a response timeout.
- React Native BLE link timeout is covered separately from command response timeout.

## Verification

- [x] `yarn check-versions`
- [x] `yarn agent:check --profile commit`
- [x] `yarn agent:check --profile pr`
- [x] Full lint passed.
- [x] Full tests passed.
- [x] Full build passed.
- [x] Firmware install polling tests: 19/19 passed.

## Release

- Version: `1.2.1`
- npm dist-tag: `latest`
- Publish workflow: `publish-npm-packages`
- Stable GitHub release/tag: `v1.2.1`


ByteZhang1024 (approved):

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -374,6 +374,15 @@ const isProtocolV2FirmwareStatusEndpointUnavailable = (error: unknown) => {
   );
 };
 
+const isProtocolV2InstallResponseTimeout = (error: unknown) => {
+  if (!error || typeof error !== 'object') return false;
+  const { code } = error as { code?: number | string };
```
originalix, line 379: P1: [BLE timeout is misclassified as install disconnect]

Native BLE transports create response timeouts as `HardwareErrorCode.BleTimeoutError` on the `errorCode` field, with the default message `The connection has timed out unexpectedly.`. This helper only reads `code` and looks for `response timeout` text, so that timeout is not excluded. Because `isProtocolV2DeviceDisconnectedError` treats the canonical timeout message as a disconnect, the install flow can carry false disconnect evidence across reconnect and accept stale finished status records as a successful firmware install.

Recognize the `errorCode` field, including `HardwareErrorCode.BleTimeoutError`, in this classifier and add a regression test using the default timeout error shape so only a genuine device disconnect establishes install evidence.

originalix:

@wabicai

Reviewed commit `43d2856e593a`. This report always reflects the latest reviewed changes.

### Review summary

The PR updates the Protocol V2 Pro2 BLE firmware-install state machine to carry disconnect evidence from the install request into reconnect/status polling, reset stale completion snapshots after link loss, and apply the same completion checks to all firmware targets. It also synchronizes all workspace package manifests to version 1.2.1.

**What needs attention:** Correct the BLE timeout classification before relying on disconnect evidence to accept finished firmware status records.

### Issues to address

- **P1 · BLE timeout is misclassified as install disconnect**
  - **Impact:** Native BLE transports represent a response timeout as HardwareErrorCode.BleTimeoutError with errorCode and the canonical message "The connection has timed out unexpectedly." The new helper reads code instead of errorCode and only matches response-timeout wording, while the existing disconnect predicate treats that canonical message as a disconnect. The timeout can therefore set protocolV2InstallDisconnectObserved, letting stale finished status snapshots after reconnect satisfy the completion rule and report firmware installation success without evidence that the request was accepted.
  - **Suggested action:** Update the install-timeout classifier to recognize the HardwareError errorCode field, including HardwareErrorCode.BleTimeoutError, while preserving the separate handling for genuine BleDeviceDisconnected errors; add a regression test using ERRORS.TypedError(HardwareErrorCode.BleTimeoutError) without a custom response-timeout message.

### Validation gaps

`yarn check-versions` passed and confirmed all 30 packages use `1.2.1`. The focused firmware-update Jest command could not run because the `jest` executable is unavailable in this workspace; CI or a bootstrapped checkout should run the new polling tests, including the default native BLE timeout shape.

PR #901, opened 2026-08-25
Chore/trezor vendor sync v26.7.4

ByteZhang1024:


packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
```diff
@@ -895,7 +949,33 @@ export class LedgerConnectorBase implements IConnector {
...
+  async configure(config: ConnectorConfig): Promise<void> {
+    const nextUrl = config.ledgerGenuineCheckWebSocketUrl;
+    if (nextUrl) {
+      let parsed: URL;
+      try {
+        parsed = new URL(nextUrl);
+      } catch {
+        throw new Error('Ledger genuine-check relay URL is invalid');
+      }
+      if (parsed.protocol !== 'wss:') {
```
originalix: P1: [Ledger genuine-check relay URL is not host-allowlisted]

When a caller supplies any syntactically valid `wss:` URL, this check accepts it and `addConfig({ webSocketUrl: ... })` sends the Ledger genuine-check session to that origin. That allows device attestation data and the genuine-check transcript to leave the intended OneKey relay and makes the result depend on an untrusted service.

Validate an SDK-owned allowlist of exact relay origins and path policy here (including rejecting userinfo, query, fragment, unexpected ports, and non-production hosts) before rebuilding the DMK.

packages/hwk-trezor-transport/src/transports/bridge.ts
```diff
@@ -191,20 +178,11 @@ export class BridgeTransport extends AbstractTransport {
     }
 
     private getProtocol(customProtocol?: TransportProtocol) {
```
originalix, line 187: P1: [Bridge protocol capability negotiation was removed]

After `init` no longer reads `protocolMessages`, `getProtocol` always selects Protocol V1 and `getRequestBody` always emits the structured protocol-message JSON body. A legacy trezord bridge that does not advertise protocol-message support will receive a format it cannot parse, so existing bridge-backed calls fail.

Restore the capability-dependent legacy framing when `protocolMessages` is absent and add tests for both legacy and current bridge responses.

packages/hwk-trezor-adapter/src/deviceAuthenticity/verifyAuthenticityProof.ts
```diff
@@ -0,0 +1,239 @@
...
+  if (caPubKeyBlacklist.includes(caPubKey)) {
+    return {
+      valid: false,
+      caPubKey,
+      rootPubKey: rootPubKeyMatch,
+      deviceCertPubKey,
+      error: 'CA_PUBKEY_BLACKLISTED',
+    };
+  }
+  const caCertValidityFrom = caCert.tbsCertificate.validity.from.getTime();
```
originalix: P2: [Trezor attestation accepts expired certificates]

This code rejects a CA whose `notBefore` is in the future but never checks `notAfter` for the CA or the device leaf. Once either certificate expires, the chain can still return `valid: true` and produce a trusted device ID.

Require the current time to fall within both certificates' validity intervals before accepting the proof, and add expired-CA and expired-leaf test vectors.

originalix:

@ByteZhang1024

Reviewed commit `5311691511af`. This report always reflects the latest reviewed changes.

### Review summary

This PR vendors the Trezor transport-common and browser layers, refreshes protobuf and protocol definitions, adds streamed Trezor device-authenticity verification, and adds Ledger genuine-check relay configuration while updating package wiring and examples.

**What needs attention:** Remove the newly introduced uncontrolled network and logging paths, then rerun browser, transport, and package validation.

### Issues to address

- **P1 · WebUSB initialization allows an arbitrary sessions-background fetch**
  - **Impact:** `sessionsBackgroundUrl` is caller-controlled and `init()` issues a `fetch` to it without origin or path validation. This expands the SDK's outbound network surface and permits arbitrary hosts to receive client requests during transport startup.
  - **Suggested action:** Restrict the URL to the same-origin generated worker asset (or a tightly defined allowlist), reject external origins, and use credentials omission for the asset probe.
- **P1 · Protocol errors log raw transport payload bytes**
  - **Impact:** When decoding fails, the new transport utility writes the first response buffer as hexadecimal to `console.warn`. Device responses can contain transaction/signing material and other sensitive protocol data, so malformed input or device errors can expose it in browser or application logs.
  - **Suggested action:** Remove the payload dump and log only non-sensitive metadata such as message type, length, and error code.
- **P1 · Ledger genuine-check adds an unapproved outbound WebSocket path**
  - **Impact:** Calling `getDeviceGenuineCheck` constructs Ledger's DMK action, which opens the documented `wss://scriptrunner.api.live.ledger.com/update/genuine` backend by default. This is a new external network path carrying an attestation transcript and device metadata outside the repository's approved network surface.
  - **Suggested action:** Route the operation through a validated OneKey-owned relay or obtain an explicit network-policy exception with enforced endpoint and data-flow controls before exposing this API.

### Validation gaps

The focused package build could not run because the workspace lacks the `tsup` executable (`/bin/sh: tsup: command not found`). Browser bundling, transport integration, and real-device attestation paths therefore still need to be validated after dependencies are installed.

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
originalix: P1: [Ledger genuine-check relay URL is not host-allowlisted]

When a caller supplies any syntactically valid `wss:` URL, this check accepts it and `addConfig({ webSocketUrl: ... })` sends the Ledger genuine-check session to that origin. That allows device attestation data and the genuine-check transcript to leave the intended OneKey relay and makes the result depend on an untrusted service.

Validate an SDK-owned allowlist of exact relay origins and path policy here, including rejecting userinfo, query, fragment, unexpected ports, and non-production hosts, before rebuilding the DMK.

packages/hwk-trezor-transport/src/transports/bridge.ts
originalix, line 187: P1: [Bridge protocol capability negotiation was removed]

After `init` no longer reads `protocolMessages`, `getProtocol` always selects Protocol V1 and `getRequestBody` always emits the structured protocol-message JSON body. A legacy trezord bridge that does not advertise protocol-message support will receive a format it cannot parse, so existing bridge-backed calls fail.

Restore capability-dependent legacy framing when `protocolMessages` is absent and add tests for both legacy and current bridge responses.

packages/hwk-trezor-adapter/src/deviceAuthenticity/verifyAuthenticityProof.ts
originalix: P2: [Trezor attestation accepts expired certificates]

This code rejects a CA whose `notBefore` is in the future but never checks `notAfter` for the CA or the device leaf. Once either certificate expires, the chain can still return `valid: true` and produce a trusted device ID.

Require the current time to fall within both certificates' validity intervals before accepting the proof, and add expired-CA and expired-leaf test vectors.

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
```diff
@@ -847,6 +902,56 @@ export class LedgerConnectorBase implements IConnector {
           ctx.clearCanceller(sessionId);
         }
       }
+      case 'getDeviceGenuineCheck': {
```
originalix, line 905: P1: [Ledger authenticity adds an unapproved outbound WebSocket path]

Calling `getDeviceGenuineCheck` now starts Ledger's DMK secure-channel networking, using either the external Ledger endpoint or a configured relay. This sends the device-attestation transcript outside the SDK's previously approved network boundary and makes the genuine result depend on an external service.

Move this exchange behind an explicitly approved backend/transport boundary, or gate it behind a narrowly fixed product-level endpoint policy before exposing the method.

packages/hwk-trezor-adapter/src/deviceAuthenticity/index.ts
```diff
@@ -0,0 +1,172 @@
...
+        return { verified: false, error: tropic.error };
+      }
+      usedDebugKey = usedDebugKey || matchedDebugKey(config, deviceModel, tropic.rootPubKey);
+    }
+
+    const deviceId = Buffer.from(
+      sha3Hash(Uint8Array.from(Buffer.from(optiga.deviceCertPubKey, 'hex')))
+    ).toString('hex');
+
+    return {
```
originalix, line 155: P1: [Safe 7 can be marked verified without the required MCU proof]

For T3W1, the streamed response may include `mcu_certificates` and `mcu_signature`, but this path requires only Optiga and Tropic before returning `verified: true`. The current Trezor policy also requires MCU/ML-DSA verification and cross-layer root, CN, and serial consistency, so a device assembled from mismatched genuine components can receive the same public result as a fully attested Safe 7. The demo already presents this boolean as “Verified.”

Implement the complete capability-based Safe 7 policy, including ML-DSA-44 MCU and cross-layer consistency checks, before allowing this branch to return `verified: true`.

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
```diff
@@ -847,6 +902,56 @@ export class LedgerConnectorBase implements IConnector {
...
+          const kit = await ctx.importLedgerKit('@ledgerhq/device-management-kit');
+          const action = (
+            dmk as unknown as {
+              // Loose return: deviceActionToPromise narrows the observable shape.
+              // eslint-disable-next-line @typescript-eslint/no-explicit-any
+              executeDeviceAction(args: { sessionId: string; deviceAction: unknown }): any;
+            }
+          ).executeDeviceAction({
+            sessionId,
+            deviceAction: new kit.GenuineCheckDeviceAction({ input: {} }),
```
originalix, line 923: P1: [Ledger genuine check bypasses the SDK network boundary]

Calling this method makes Ledger DMK open an external WebSocket to its default service or the configured OneKey relay. The hostname validation narrows the relay destination, but the connector still sends the device-attestation transcript through a new outbound path and delegates the authoritative verdict to that service, expanding the SDK's network and trust boundary.

Move the DMK genuine-check WebSocket exchange into an explicitly approved host-owned integration injected into the adapter, and keep this connector limited to device communication.

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
originalix, line 905: P1: [Ledger genuine-check adds an unapproved outbound WebSocket path]

Calling `getDeviceGenuineCheck` constructs Ledger's DMK genuine-check action, which opens `wss://scriptrunner.api.live.ledger.com/update/genuine` by default. This sends the attestation transcript and device metadata through a new external network path outside the repository's approved network surface.

Route the operation through a validated OneKey-owned relay, or complete an explicit network-policy review with enforced endpoint and data-flow controls before exposing this API.

ByteZhang1024: Intentional by design. Direct-to-Ledger is an approved destination (Ledger's own official manufacturer attestation service — same endpoint every Ledger integration uses). The only alternative destination, a caller-supplied relay via configure({ledgerGenuineCheckWebSocketUrl}), is already restricted to OneKey-owned domains by assertAllowedLedgerRelayUrl (LedgerConnectorBase.ts:139-171). No third destination is reachable from either path.

packages/hwk-trezor-transport/src/transports/bridge.ts
ByteZhang1024: Fixed in 9e90aa79c9

ByteZhang1024: Fixed in 9e90aa79

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
ByteZhang1024: Not unique to getDeviceGenuineCheck — installApp (InstallOrUpdateAppsDeviceAction) and listAvailableApps (ListAvailableAppsDeviceAction) already run through the identical dmk.executeDeviceAction() path and equally reach Ledger's own manager-api backend over the network (see DeviceApps.ts:100 — the code itself distinguishes offline-APDU actions from manager-api-backed ones). This connector wraps Ledger's DMK, whose entire purpose is abstracting over exactly this online/offline distinction per device-action; it was never a network-free layer. Carving out only the genuine-check call would not establish the invariant this finding asks for while leaving the other two DMK-network operations untouched in the same class — not doing this refactor.

packages/hwk-trezor-adapter/src/deviceAuthenticity/index.ts
ByteZhang1024: Correct but intentional — see the verified field's own doc comment: it's a UX preview only, not the reward-issuance check. Reward claiming is fully server-side now (85a39971a), which independently verifies the raw proof including MCU. No fix needed here.

packages/hwk-ledger-adapter/src/connector/LedgerConnectorBase.ts
ByteZhang1024: By design. Direct-to-Ledger is Ledger's own official verification service. The relay override is already locked to OneKey domains via assertAllowedLedgerRelayUrl. Not a gap.

originalix (approved):

PR #902, opened 2026-08-26
perf: optimize Protocol V2 BLE transfer throughput and metrics

wabicai:
## Summary

- Optimize Protocol V2 BLE writes across React Native, Electron, and CLI with negotiated packet capacity, write-without-response support, and explicit fallback behavior.
- Avoid repeating Main PIN selection when a Protocol V2 device is already unlocked in the standard wallet, while still forcing recovery when the cached session resolves to another wallet.
- Use the verified 1960-byte chunk budget for fixed firmware and wallpaper package paths while keeping generic filesystem writes conservative.
- Improve lossless Pro2 host-asset LZ4 compression with a firmware-safe fallback for incompressible input.
- Report whole-batch firmware transfer bytes, average bytes per second, and elapsed time across resources, components, verification, recovery, and retry delays.
- Preserve the original transport error cause so App-side firmware failure classification is not swallowed by retry wrappers.
- Make Protocol V2 firmware cancellation retryable: abort all queued requests when no connectId is supplied, stop file-transfer recovery after abort, and avoid sending the unsupported protocol Cancel command while the device is in bootloader mode.

## Published package

- Published version: `1.2.2-alpha.7`
- npm dist-tag: `next`
- Publish workflow: https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/33386963375
- Companion App PR: https://github.com/OneKeyHQ/app-monorepo/pull/13025

## Compatibility decision

- `LowlevelTransportSharedPlugin` is an internal adapter surface in the supported OneKey runtimes; there is no supported third-party callback-free BLE plugin compatibility requirement.
- The 192-byte Protocol V2 lowlevel fallback is therefore intentional.
- Updated adapters can report negotiated capacity explicitly, capped at 244 bytes; the CLI uses that path.
- Protocol V1 behavior, dual-protocol probing, WebUSB limits, and serialized Protocol V2 calls remain unchanged.

## Measured results

- Pre-optimization reference: wallpaper with 64-byte transport packets / 1800-byte file chunks measured 8.81 KiB/s.
- Optimized wallpaper with 244-byte negotiated packets / 1960-byte fixed-path chunks measured 16.72 KiB/s.
- Pro2 4B8F bootloader update (2,440,562 bytes): 145.62 s transfer, 16.37 KiB/s average, 13.65 s install, 211.96 s CLI end-to-end, 0 retries.
- A 15 ms firmware connection interval measured 20.3 KiB/s versus 14.22 KiB/s at 30 ms. Connection-parameter changes remain outside this SDK PR.

## Validation

- `yarn agent:check --profile commit` passed.
- Core test and build passed.
- Lowlevel transport test and build passed.
- Focused Protocol V2 tests: 318 passed, 4 skipped.
- Focused wallet-session tests: 76 passed.
- Focused lowlevel tests: 20 passed.
- `yarn check-versions` passed for all 32 packages at `1.2.2-alpha.7`.
- Physical Desktop WebUSB cancellation/retry and transfer metrics were verified with Neo/Pro2 Protocol V2.

## Test plan

- [x] Verify BLE packet capacity negotiation and the intentional 192-byte lowlevel fallback.
- [x] Verify an already-unlocked standard wallet does not repeat Main PIN selection and mismatched cached sessions still recover through Main PIN.
- [x] Verify fixed-path firmware/wallpaper chunking and host-package round trips.
- [x] Verify whole-batch firmware bytes, rate, and elapsed time.
- [x] Verify cancellation aborts the active transfer without recovery or wrapper replacement.
- [x] Verify a bootloader-mode retry reuses the active Protocol V2 workflow.
- [x] Publish `1.2.2-alpha.7` through GitHub Actions and verify npm availability.


originalix:

@wabicai

Reviewed commit `9ed2a6893b26`. This report always reflects the latest reviewed changes.

### Review summary

Before: Protocol V2 BLE probing and file transfer used conservative fixed packet/chunk sizes, while firmware cancellation registered a protocol Cancel action.
After: BLE transports derive packet capacity from negotiated MTU/plugin capability, fixed-path firmware and wallpaper uploads use larger validated chunks, host asset packages use adaptive LZ4 blocks, and V2-only methods can probe directly. Firmware transfer progress now reports aggregate timing/rate, and package versions are aligned to 1.2.2-alpha.7.

Adjacent behavior intentionally preserved: generic long-path file writes remain at the conservative 1800-byte BLE limit, Protocol V1 probing/framing is unchanged, wallet-session recovery avoids redundant Main PIN prompts, and no new network surface was added.

**What needs attention:** Verify that cancelling a bootloader firmware install tears down or otherwise invalidates the active transport, then run physical-device cancellation and reconnect tests on React Native, Electron, and CLI BLE paths.

### Issues to address

- **P2 · Firmware update cancellation leaves the bootloader link alive**
  - **Impact:** At both install-request setup and status polling, the new cancelable action resolves immediately. Device.interruptionFromUser therefore skips the fallback disconnect branch, while DeviceCommands.cancel()/transport.cancel() do not itself release the lowlevel or React Native Protocol V2 link. A cancellation can leave the bootloader session and in-flight status call alive until a later timeout or cleanup, causing a stuck device and delayed reconnect.
  - **Suggested action:** Replace the no-op action with an explicit link invalidation/disconnect (or clear the action before invoking the existing cancellation fallback), and add a test asserting cancellation releases the active V2 transport.

### Validation gaps

`yarn agent:check --profile pr` passed context, whitespace, and package-version checks but could not run full lint because `eslint` is not installed. Focused Jest execution was also unavailable because the workspace has no `jest` binary. Physical-device tests are still needed for cancellation teardown, MTU edge values, and negotiated packet sizes on each BLE platform.

packages/hd-transport-lowlevel/src/index.ts
```diff
@@ -26,7 +26,8 @@ const { check, ProtocolV1, parseConfigure } = transport;
 
 const PROTOCOL_PROBE_TIMEOUT_MS = 1000;
 const PROTOCOL_V2_PROBE_TIMEOUT_MS = 5000;
-const LOWLEVEL_PROTOCOL_V2_PACKET_LENGTH = 64;
+const LOWLEVEL_PROTOCOL_V2_PACKET_LENGTH_FALLBACK = 192;
```
originalix, line 29: P1: [Preserve the lowlevel packet-size fallback for existing plugins]

When an existing `LowlevelTransportSharedPlugin` does not implement the new optional `getProtocolV2PacketCapacity`, this fallback makes the shared writer send 192-byte packets directly to `plugin.send()` instead of the previous 64-byte chunks. Native plugins that forward each call to a BLE characteristic with a smaller maximum no-response write can reject or lose the V2 probe/request, causing connection, upload, or firmware-transfer timeouts even though they still satisfy the public plugin contract.

Please keep 64 bytes as the callback-free compatibility fallback and let the updated Noble CLI opt into 192/244 bytes through `getProtocolV2PacketCapacity`; add a regression test for a plugin that omits the callback.

wabicai: 忽略，目前不存在第三方 lowlevel 接入。192 经过验证是符合预期的

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -3174,7 +3199,9 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
     this.protocolV2InstallTerminalSuccessObserved = false;
     const commands = this.device.getCommands();
     await commands.typedCall('DeviceFirmwareUpdateStage', 'Success', { targets });
-    this.device.setCancelableAction(() => this.device.getCommands().cancelDevice());
+    // Bootloader does not register the protocol Cancel message. Keep the
+    // operation locally cancellable; interruptionFromUser() disposes the link.
+    this.device.setCancelableAction(() => Promise.resolve());
```
originalix, line 3209: P2: [Firmware update cancellation leaves the bootloader link alive]

When the install request is staged, this callback is registered as the active cancelable action, but it resolves without invalidating or disconnecting the Protocol V2 link. `Device.interruptionFromUser()` therefore skips its fallback disconnect path, and `DeviceCommands.cancel()`/`transport.cancel()` do not tear down the lowlevel or React Native link. A user cancellation can leave the bootloader status call and session alive until a later timeout or cleanup, delaying reconnect and potentially leaving the device stuck.

Make the cancelable action explicitly invalidate/disconnect the active link (or clear it so the existing disconnect fallback runs), and add a cancellation-then-reacquire regression test.

wabicai: ignore

wabicai: ignore

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/hd-transport-lowlevel/src/index.ts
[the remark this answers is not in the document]
wabicai: ignore

packages/core/src/core/index.ts
```diff
@@ -1333,7 +1341,10 @@ export const cancel = (context: CoreContext, connectId?: string) => {
       );
       // Abort before rejecting: rejectRequest releases the task and would make
       // its AbortController unreachable to an in-flight method loop.
-      requestQueue.abortRequestsByConnectId(connectId);
+      // This branch rejects every queued request below. Abort the same set first so
+      // methods whose physical connectId is selected internally (for example
+      // Desktop WebUSB firmwareUpdateV4) cannot keep retrying after rejection.
+      requestQueue.abortAllRequests();
```
sidmorizon: **Device cancellation stops unrelated operations**

Severity: `non-severe`

Calling `cancel(connectId)` runs `abortAllRequests()` for every active device. Concurrent signing or firmware updates on other devices can stop unexpectedly.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2234,18 +2253,22 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
         });
         return processedSize + source.size;
       } catch (error) {
+        this.throwIfAborted();
```
sidmorizon, line 2261: 🟠 P1: Cancel-all resumes WebUSB firmware writes

Severity: `severe`

After `cancel()` without a connect ID, `throwIfAborted()` sees an active signal on WebUSB. Failed transfers can reconnect and resume after cancellation.

packages/core/src/api/FirmwareUpdateV4.ts
```diff
@@ -2225,7 +2244,7 @@ export default class FirmwareUpdateV4 extends FirmwareUpdateBaseMethod<FirmwareU
                 transferredBytes,
                 totalBytes: totalSize,
                 rateBytesPerSecond:
-                  elapsedMs > 0 ? Math.round((chunkEnd / elapsedMs) * 1000) : undefined,
+                  elapsedMs > 0 ? Math.round((transferredBytes / elapsedMs) * 1000) : undefined,
```
sidmorizon, line 2252: **Retry regresses batch transfer progress**

Severity: `non-severe`

When a file restarts after recovery, `transferredBytes` returns to its beginning. The public progress stream can jump backward from its last confirmed value.

packages/connect-examples/expo-example/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "expo-example",
-  "version": "1.2.2-alpha.6",
+  "version": "1.2.2-alpha.8",
```
sidmorizon: **Examples still install the old SDK**

Severity: `non-severe`

The examples declare version `alpha.8` but pin OneKey dependencies to `alpha.6`. Their builds can bypass every SDK change in this release.

packages/core/src/protocols/protocol-v2/walletSession.ts
```diff
@@ -387,8 +387,10 @@ export async function getProtocolV2WalletSession(
       standardWalletSelected = true;
     } else if (!standardWalletSelected) {
       // Without passphrase protection there is no empty-passphrase selector.
-      // Main PIN selection is the only authoritative switch back to the standard wallet.
-      await selectMainPin(true);
+      // An unlocked non-Attach-PIN device is already in the only available wallet context.
+      // Force Main PIN only when recovering from a mismatched cached standard session.
+      await selectMainPin(forceMainPin);
+      standardWalletSelected = true;
```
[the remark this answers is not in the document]
wabicai: Fixed in 9dc765a64. The onlyMainPin path now refreshes authoritative DeviceStatus before deriving mainPinAuthenticated unless mainPinSelected is explicitly authoritative. The useEmptyPassphrase caller also rejects an Attach-PIN wallet for the main-wallet path, with regression coverage for stale cached features.

packages/core/src/core/index.ts
wabicai: Fixed in 9dc765a64. cancel(connectId) now aborts only matching request controllers through abortRequestsByConnectId(connectId); matching covers method.connectId, device.mainId, and device.getConnectId(). Requests for other devices remain active, and focused lifecycle tests cover the scoped behavior.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Fixed in d4ae60c95. cancel() without a connect ID now aborts every request controller before either BLE or WebUSB queue rejection, so FirmwareUpdateV4.throwIfAborted() stops recovery/reconnect loops on all transports. Added a WebUSB cancel-all regression test that verifies the signal is aborted before the task is released.

packages/core/src/api/FirmwareUpdateV4.ts
wabicai: Fixed in d4ae60c95. Protocol V2 transfer progress now maintains a batch-level transferred-byte watermark and suppresses lower offsets emitted while a restarted file catches up. The new regression test forces a retry after 2 KB and verifies the public stream remains [1000, 2000, 3000].

packages/connect-examples/expo-example/package.json
wabicai: Fixed in d4ae60c95 and carried forward by d20768b7f. The Expo, playground, and Electron examples were first aligned with the current SDK, then Lerna advanced both their package versions and internal OneKey dependency pins to 1.2.2-alpha.10. The stale alpha.6 lock entries were removed.

packages/hd-transport-lowlevel/src/index.ts
sidmorizon, line 29: **Legacy low-level plugins lose safe fallback**

Severity: `non-severe`

When a plugin omits `getProtocolV2PacketCapacity`, the fallback grows from 64 to 192 bytes. Existing 64-byte adapters then fail every Protocol V2 call.

Suggested fix:

```ts
const LOWLEVEL_PROTOCOL_V2_PACKET_LENGTH_FALLBACK = 64;
```

originalix (approved):

PR #918, opened 2026-09-05
fix: dispose Electron BLE before Node teardown (OK-62081)

wabicai:
Quitting Electron after BLE initialization can leave Noble native managers alive until Node environment cleanup, where releasing their ThreadSafeFunctions can deadlock the macOS main thread. Both BLE transports now support bounded process disposal while Node is still running, with one native stop per shared instance.

- OneKey unregisters IPC, cancels initialization, scans and native connects, drains connection cleanup, and guards callbacks after native release.
- Trezor tracks native connect settlement and every Noble instance created during adapter recovery. Renderer disposal preserves native managers; process disposal releases them.
- Hosts can defer native release until both transports finish and deduplicate shared instances.
- Renderer replacement cancels and drains old OneKey enumerations, clears their timers, and prevents late scan callbacks or delayed Bluetooth initialization from starting work for a destroyed window.

Desktop integration: https://github.com/OneKeyHQ/app-monorepo/pull/13194
Issue: https://onekeyhq.atlassian.net/browse/OK-62081

Validation:
- Focused OneKey suite: 24 tests passed, including enumeration retirement, delayed scan callbacks, power-on after renderer destruction, and native connect cancellation/release ordering.
- SDK commit and PR checks passed, including version consistency, lint, tests, and package builds.
- The shared-Noble reproduction passes against the locally built fix: the replacement Trezor scan remains active past the old enumeration deadline and five subsequent polls.

All 32 SDK packages and internal references are aligned to `1.2.2-alpha.104`. The 24 public SDK packages and the existing workflow's `hardware-example` package (25 total) were published with dist-tag `next` from `67835939d9f779ffd50cd8b482d8c5b53bd580bd` via https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/33969355609. This release includes the follow-up enumeration-retirement fix `1985d3b89`; each package's version, dist-tag and gitHead were verified. No new native patch is included.

Signed TestFlight/MAS validation with connected hardware remains necessary; simulated callback and Electron cleanup tests do not replace device QA.


packages/hd-transport-electron/src/noble-ble-handler.ts
```diff
@@ -2207,3 +2280,65 @@ export function setupNobleBleHandlers(webContents: WebContents): void {
...
+      clearTimeout(timeout);
+      cleanupNobleListeners();
+      if (instance && persistentStateListener) {
+        instance.removeListener('stateChange', persistentStateListener);
+        persistentStateListener = null;
+      }
+      for (const id of deviceIds) cleanupDevice(id);
+      discoveredDevices.clear();
+      directConnectCooldownUntil.clear();
+      if (instance) releaseNoble(instance);
```
originalix, line 2426: P1: [Drain in-flight connects before stopping Noble]

When app quit overlaps a pending `peripheral.connect`/`connectAsync`, the shutdown code only rejects the JavaScript-facing waiter; it does not natively cancel or await the underlying connect. This line can therefore call `stop()` first, after which a late connect callback resumes and calls disconnect or emits through a released native manager, risking a process-shutdown crash.

Please track and cancel the native connect operation (using Noble’s `cancelConnect` path where supported) and include its settlement in the disposal barrier before invoking `releaseNoble`. The Trezor handler’s `_connectAttempts`/`disposeForAppQuit()` path needs the same ordering guarantee.

wabicai: Fixed in [4f41c41f3](https://github.com/OneKeyHQ/hardware-js-sdk/commit/4f41c41f3c5ca670731f01e661387f80f79ab26f), published as `1.2.2-alpha.103` via [GitHub Actions](https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/33964938926).

Both OneKey and Trezor now track the underlying native connect operation separately from the JavaScript-facing waiter, call the native `cancelConnect` path where available, and include native settlement and late-success disconnect cleanup in the disposal barrier before `releaseNoble`. Cleanup remains bounded at 3.5 seconds; if it times out, the native-release guard prevents late connect callbacks from disconnecting through the released manager.

Validation: focused regression tests cover pending-connect cancellation, settlement before release, and late callbacks after the disposal timeout for both transports. The App now consumes `.103` in [813d37fb1](https://github.com/OneKeyHQ/app-monorepo/commit/813d37fb1fffb3f5e10f9ceaaf906fd721febfc4); six Electron exit scenarios against the installed npm packages passed.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-bb360e07-82fa-5966-847f-0f088c8cfe04)

originalix (approved):

PR #919, opened 2026-09-05
fix: stabilize device cleanup and pace Pro2/Neo USB signing

wabicai:
Cancelling a BLE operation or immediately starting another call could overlap old connection setup or teardown. Repeated calls and SDK resets also left request records and React Native BLE resources behind.

Requests are registered before queue waits, and cancellation uses the existing request signal and per-device cleanup barrier. Device cancellation captures the old commands and run state and is deduplicated per connection attempt. Completed V1/V2 calls preserve their acknowledged result while registering the actual run cleanup with that same barrier, so the next acquire waits for release. A cleanup deadline does not make an unfinished connection reusable.

Pro2 and Neo signing calls over browser WebUSB, desktop WebUSB, and Node USB now append a one-second cooldown to that per-connectId post-call barrier. The current signing result is returned at the existing time; only the next device call waits. BLE, older device families, non-signing methods, and other devices remain unaffected. The same cooldown is registered after success, cancellation, or a signing error so an immediate retry cannot race device recovery.

React Native V1 cancellation rejects its pending read and waits for native disconnect. SDK disposal now reaches Core and drains the transport's scans, scan timers, pending reads and owned native connections. Ordinary disposal leaves the shared ble-plx manager available to other connectors. A manager reset waits for asynchronous singleton destruction before another transport can acquire; destruction failures or deadlines do not permit reuse of the old manager. Both SDK entries remove request records when calls settle, including out-of-order failures.

Connection setup failures stop the outer poll. Existing scan/connect/GATT timing logs remain available; scan windows, connection retry parameters and transfer packet sizes are unchanged.

Release target:
- All workspace package versions and exact internal dependencies are aligned at `1.2.2-alpha.106`.
- After merge and explicit publication approval, publish the package set with npm dist-tag `next`.
- The companion App PR will then consume `1.2.2-alpha.106`.

Validation:
- `yarn agent:check --profile pr` passed on the release commit.
- `yarn agent:check --profile commit` passed on the release commit.
- `yarn build` passed for all 33 packages.
- `yarn check-versions` passed: all 32 checked packages are at `1.2.2-alpha.106`.
- Regressions cover V1/V2 cleanup before the next acquire, queued/repeated/late cancellation, reset waiting across transport instances, failed native destruction, scan timer cleanup, pending-read/native-disconnect settlement, SDK request/listener cleanup, and Pro2/Neo USB signing cooldown boundaries.

Companion App fix: https://github.com/OneKeyHQ/app-monorepo/pull/13196.
Related symptom: https://onekeyhq.atlassian.net/browse/OK-62115.

Physical iOS/Android/Electron cancellation, transfer performance, Pro2/Neo continuous USB signing, and post-upload stability still require device validation. This does not establish the cause of the reported iOS freeze.


packages/core/src/core/index.ts
```diff
@@ -367,42 +367,38 @@ const onCallDevice = async (
...
     const pollingId = pollingManager.start(connectId);
-    device = await ensureConnected(
-      context,
-      method,
-      connectId,
-      pollingId,
-      task.abortController?.signal
-    );
+    device = await ensureConnected(context, method, connectId, pollingId, method.abortSignal);
+    if (method.abortSignal?.aborted) {
```
originalix, line 415: P1: [Cancellation can slip past the final pre-run waits]

After this check, the call still awaits the callback and cleanup barriers. If cancellation arrives during either wait, `cancel()` aborts and removes the request task, but this flow can then continue into `Device.run`; React Native may re-enable the disposed command channel, device I/O can still execute, and `resolveRequest()` cannot settle the removed task.

Please make those post-acquire waits abort-aware and enforce a final signal check immediately before constructing the `Device.run` deferred.

packages/hd-ble-sdk/__tests__/device-state-events.test.ts
```diff
@@ -13,3 +25,64 @@ describe('hd-ble-sdk device state events', () => {
...
   });
 });
+
+const sdk = sdkExport as unknown as {
+  init(settings: Record<string, unknown>): Promise<boolean>;
+  call(params: Record<string, unknown>): Promise<unknown>;
+  dispose(): Promise<void>;
+};
+
+describe('SDK request cleanup', () => {
```
originalix, line 40: P2: [The new entry-SDK tests are not part of the test command]

These cleanup tests are added under `hd-ble-sdk`, and the equivalent suite is added under `hd-common-connect-sdk`, but neither package defines `scripts.test` or a Jest configuration. Because the root command uses `lerna run test`, both suites are skipped, so the disposal and out-of-order request changes can regress while configured checks stay green.

Please add usable Jest configurations and `test` scripts to both package manifests, then verify `yarn test` discovers these files.

originalix:
## Automated review did not complete

Review of commit `2c23b11e9ec2` did not finish successfully.

No automated review conclusion was produced for this commit.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9ff2b1c7-59c8-5afc-808a-081fb02fb54a)

packages/hd-transport-react-native/src/index.ts
```diff
@@ -503,10 +509,34 @@ export default class ReactNativeBleTransport {
...
-    if (this.blePlxManager) return Promise.resolve(this.blePlxManager);
-    this.blePlxManager = new BlePlxManager();
-    return Promise.resolve(this.blePlxManager);
+  async getPlxManager(): Promise<BlePlxManager> {
+    while (bleManagerResetPromise) {
+      await this.waitForManagerReset();
+    }
+    if (this.stopped) throw ERRORS.TypedError(HardwareErrorCode.BleDeviceDisconnected);
+    if (!this.blePlxManager) this.blePlxManager = new BlePlxManager();
+    return this.blePlxManager;
```
sidmorizon, line 518: **Existing BLE clients reuse destroyed manager**

Severity: `non-severe`

After a shared manager reset, `getPlxManager` returns another client's cached manager. [`resetPlxManager`](packages/hd-transport-react-native/src/index.ts:2209-2271) clears only its own reference, leaving existing clients bound to the destroyed singleton.

originalix (approved):

PR #903, opened 2026-08-28
fix: align Protocol V2 Cardano derivation with origin/dev

wabicai:
## Summary

Protocol V2 wallet sessions follow firmware `origin/dev` (`f83f38bb6`). `DeviceSessionAskPassphrase.seed_domains` owns derivation. `DeviceSessionGet` is read/resume only and must not send field 3.

This also restores Pro1 Attach PIN product behavior: `select-hidden` while Attach PIN is unlocked reuses the current session. It does not lock and does not open the passphrase picker.

## Intent and context

origin/dev rejects unknown `DeviceSessionGet` field 3. Cardano must be requested on AskPassphrase. Attach PIN has no host passphrase, so Cardano uses an empty host Ask on the current SE session.

`+ Hidden wallet` while Attach PIN is unlocked is not a PIN mismatch. Pro1 returns the current Attach PIN wallet. The App shows the green "Wallet exists" toast only when that wallet is already in local DB.

## Root cause

- Sending `Get.seed_domains` fails on origin/dev.
- A real passphrase Ask on an Attach PIN session starts `SESSION_NEW` / can fail on SE.
- Treating `select-hidden` + Attach PIN as `DeviceCheckUnlockTypeError` mapped to "The entered PIN does not correspond to the current wallet." That copy is wrong for this path.

## Design decisions

- **AskPassphrase** sends `[Standard]` on wallet open; Cardano sends `[Standard, Cardano]`. Never send `[]`.
- **Get** carries only `session_id` / `btc_test_address`.
- **Attach PIN Cardano**: empty host `AskPassphrase({ passphrase: '', on_device: false, seed_domains: [Standard, Cardano] })` then `Get()`. Empty host Ask does not open the passphrase UI. Firmware must GEN Cardano in place and keep `unlocked_by_attach_to_pin`.
- **Passphrase hidden + Cardano**: extra Cardano still needs a **real** passphrase Ask, even if the device is still unlocked. origin/dev memzeros the secret after GEN, so Get cannot add Icarus later. This is expected, not a lock-state bug.
- **Passphrase off**: firmware Get auto-requests Cardano; SDK may send a second `Get()` if the first response is Standard-only. Do not AskPassphrase (firmware returns PassphraseDisabled).
- **`select-hidden` + Attach PIN unlocked**: `readCurrentAttachPinSession`. Reuse the current wallet. Do not `lockDevice`. Do not show the passphrase picker.
[...]

PR #926, opened 2026-09-08
fix: backport App 6.5.2 wallet compatibility in SDK 1.2.2

wabicai:
App `release/v6.5.2` pins Hardware SDK `1.2.1`. Its Protocol V2 wallet flow sends `seed_domains` on `DeviceSessionGet`, while current firmware generates domains through `DeviceSessionAskPassphrase`. It also repeats Main PIN selection for an already unlocked standard wallet when passphrase protection is disabled.

This backport prepares **SDK 1.2.2** for the App 6.5.2 hot update with two focused fixes:

- Send `[Standard]` or `[Standard, Cardano]` on `DeviceSessionAskPassphrase`, read generated domains from `DeviceSession`, and supplement Cardano only when needed. Preserve wallet identity checks and Attach PIN context during derivation.
- Refresh live wallet status before skipping redundant Main PIN selection. Locked devices, Attach PIN contexts, and mismatched cached standard wallets retain the authentication flow.

Base: `release/app-6.5.2-sdk-1.2.1`, fixed at SDK tag `v1.2.1` (`43d2856e5`). Selected upstream wallet-session changes come from `fd5df217b` (#903), `9ed2a6893`, and `9dc765a64`; unrelated upstream changes are excluded.

The implementation changes one Core source file, three generated schema/type outputs, three existing test files, and the existing wallet-session documentation. The release preparation aligns all 30 SDK package versions and their existing 1.2.1 internal references to 1.2.2, following the previous release's manifest scope, with the four existing example manifests kept in sync. Public wallet parameters (including `deriveCardano` and `resume-hidden`), UI event contracts, Protocol V1, third-party dependencies, transport implementations, and firmware submodule pointers remain unchanged.

Protobuf outputs were regenerated with `yarn update-protobuf` using the SDK 1.2.1 firmware schema snapshot (`723bf8bab4`) with only `messages_device_session.proto` taken from firmware commit `f83f38bb6b`. This limits the generated diff to the three wallet-session field changes without moving the submodule pointer.

Validation for the 1.2.2 preparation:

- `yarn check-versions`: passed; all 30 SDK packages are at 1.2.2.
- `yarn agent:check --profile commit`: passed after building the private Trezor packages in dependency order; 2,395 tests passed across 174 suites (4 tests skipped), and all 30 package builds passed.
- `yarn agent:check --profile pr`: passed, including full repository lint, tests, and build.
- Packed all 24 non-private SDK packages locally, without publishing.
- App `release/v6.5.2` installed these local tarballs with its release dependencies; postinstall/web-embed build and `yarn agent:check --profile commit` (including App TypeScript) passed. Temporary local package resolutions were removed after validation.
- The original backport validation confirmed both focused regression cases fail against SDK 1.2.1 and pass with this patch; security review covered wallet-identity validation, missing-domain rejection, Attach PIN context, and unchanged public session/PIN event boundaries.

**SDK 1.2.2 has not been published to npm.** Publication is explicitly on hold for validation. Physical-device validation has not been performed. The companion App dependency PR, OneKeyHQ/app-monorepo#13275, remains draft until the SDK is approved and published, its registry lockfile is updated, and device validation is complete.


packages/core/src/protocols/protocol-v2/walletSession.ts
```diff
@@ -259,6 +292,26 @@ export async function getProtocolV2WalletSession(
...
+    keepAttachPin?: boolean;
+  } = {}) => {
+    await askDevicePassphrase(
+      device,
+      { passphrase: '', on_device: false },
+      deriveCardano,
+      markWalletStatusRefreshed
+    );
+    if (keepAttachPin && device.features) {
+      device.features.unlockedAttachPin = true;
```
originalix, line 309: P1: [Persist the Attach PIN logical context]

When a current Attach PIN wallet requests Cardano, `AskPassphrase` refreshes firmware status after `unlocked_by_attach_to_pin` has been cleared. This assignment writes the replacement marker through `device.features`, but `features` is a projected compatibility object, not the canonical device state.

`OpenWalletSession` then reads `getDeviceState()` and sees `unlockedAttachPin === false`, so the same Attach PIN + Cardano flow is rejected with `DeviceCheckUnlockTypeError`. Persist this logical Attach PIN context in durable session/state data that `OpenWalletSession` consumes after the status refresh, and add a public `openWalletSession` regression test for this path.

PR #931, opened 2026-09-11
fix: use release versions and commits for changelogs

wabicai:
Release notes for v1.2.1 compared against v1.1.32 and omitted its direct BLE fixes because GitHub's automatic notes selected a Git ancestor and listed PRs. Select the highest lower published stable version and generate notes from the commits between the release tags, excluding patch-equivalent commits already released before a squash merge or cherry-pick.

Validation:
- actionlint, YAML parsing, and shell/Python syntax checks
- Eight focused regressions, including replaying v1.2.0 → v1.2.1, numeric version ordering, maintenance releases, drafts/prereleases, pagination, first release, and a missing baseline tag
- `yarn agent:check --profile pr` (full lint, tests, build, and package version checks)

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -129,6 +129,60 @@ jobs:
             echo "✅ Release $TAG_NAME does not exist, will create"
           fi
 
+      - name: Generate release notes
+        if: steps.check-release.outputs.release_exists == 'false'
+        env:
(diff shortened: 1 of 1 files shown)


wabicai:
This changelog baseline fix is now included in https://github.com/OneKeyHQ/hardware-js-sdk/pull/932 (`7b38273b9`). Landing it there so the release workflow change ships with the SDK transport fixes.

PR #932, opened 2026-09-11
fix: stabilize BLE MTU negotiation and WebUSB discovery

wabicai:
BLE connection setup can stall during MTU negotiation or reuse an unhealthy link after reconnect. This change integrates Android MTU setup and BLE recovery from #933 and #935, together with the existing WebUSB discovery coordination.

- On Android, separate the native connection timeout from MTU negotiation. Normally negotiate MTU before service discovery, with a 12-second limit; after an explicit GATT cache refresh, let discovery finish first.
- Drop links whose MTU negotiation times out or remains at the default MTU, then wait 5 seconds before retrying. Repeated setup failures reach the manager-reset guard.
- Refresh Android GATT caches after firmware reconnects or stale-table symptoms, and provide a bounded wake attempt for a fully silent Classic device.
- Preserve the iOS MTU implementation from #935 and Electron BLE connection reuse on routine cancellation.
- Coordinate WebUSB discovery with active requests and connection setup, preserve initialization errors, and retain the V1 USB retry behavior.
- Generate release notes from published stable versions and commits between tags, including #931.

Published `1.2.2-alpha.118` to npm with the `next` dist-tag from commit `f73c3ec61e0ffe97c1d38b18058a18e169a72217`. The release commit changes package versions and internal version references only; communication code is unchanged from the integrated #935 implementation.

Validation: `yarn agent:check --profile pr` and `yarn agent:check --profile commit` passed in an isolated worktree with its own locked dependencies. Full lint, tests, and builds passed. Core: 88 suites / 1317 tests passed, 4 skipped. React Native BLE: 203 tests passed. WebUSB/Electron transport: 70 tests passed. The [npm publication workflow](https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/34819357509) successfully built and published 25 packages, including the existing example package.

Physical iOS/Android verification of the final cleanup changes listed in #935 is still pending.

Jira: [OK-62080](https://onekeyhq.atlassian.net/browse/OK-62080).

[OK-62080]: https://onekeyhq.atlassian.net/browse/OK-62080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #905, opened 2026-08-31
v1.2.2-alpha.101 fix: recover Attach PIN sessions and surface native BLE bond errors

wabicai:
## Summary

This PR completes two related hardware flows: wallet-state reconciliation for Protocol V2 devices and typed Bluetooth pairing errors across the SDK boundary.

### Wallet and PIN state

- Recover an expired Protocol V2 hidden-wallet session from the wallet selected through Attach PIN, then validate its passphrase state before accepting it.
- Keep Main PIN, Attach PIN, hidden-wallet, deprecated-wallet, reset-wallet, and selected-wallet transitions isolated.
- Keep standard hardware wallets in deprecated-wallet reconciliation; exclude only QR-based wallets.
- Avoid redundant Main PIN or passphrase prompts after Attach PIN recovery.

### Bluetooth error contract

- Keep `@stoprocent/noble` pinned to `2.3.16`.
- Forward native macOS CoreBluetooth error domain, code, and description from connection, disconnect, and notification-subscription callbacks.
- Track initial, forced-reset, and fresh-scan Noble connections in one cancellable generation state, so the Core 60-second deadline can terminate them without late-callback revival.
- Preserve typed error metadata through Electron IPC with JSON-safe serialization, including legacy preloads that still forward raw error envelopes, with no localized-text heuristic.
- On Android, establish GATT before system bonding and close the native connection when bonding fails or is cancelled.
- Keep hd-core and HWK as separate error systems; this PR adds no cross-system compatibility layer.
- Document the SDK error contract and add the shared `hardware-error-handling` skill for future changes.

## Release and integration

- Merged the latest `onekey` branch, including `03e42ae20`.
- Published `1.2.2-alpha.101` through the repository GitHub Action: https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/33839307794
- The `alpha.101` publish and package-build workflows completed successfully; the current review follow-up remains unpublished.
- App integration is updated in https://github.com/OneKeyHQ/app-monorepo/pull/13132; head `3f24654737` consumes `1.2.2-alpha.101` and carries the synchronized Noble runtime patch while this SDK follow-up remains unpublished.

## Upstream

- Proposed the generic Noble macOS error-propagation fix upstream: https://github.com/stoprocent/noble/pull/141
- The SDK/App patch remains in place until an upstream release is adopted deliberately.

## Validation

- `yarn agent:check --profile commit` passed on follow-up head `df2422fb9`.
[...]

PR #933, opened 2026-09-12
fix: recover Android BLE links and stop desktop BLE teardown on routine cancel

originalix:
Stacked on #932 (base is `fix/bound-android-ble-mtu-negotiation`, not `onekey`), so the diff here is just these changes. Bumps the line to `1.2.2-alpha.115` (`alpha.114` is already published).

Two independent BLE fixes: sections 1-3 are Android, section 4 is desktop Electron. They share no code.

## Android

Root-caused from an Android field log where a Pro 2 could not create a second account and a Classic 1S could not be onboarded at all.

## 1. Service discovery ran behind an abandoned MTU exchange

`BLE_MTU_REQUEST_TIMEOUT_MS` (3s) races `device.requestMTU` and then calls `cancelTransaction`, but that only disposes the JS subscription. Verified against the `rxandroidble 1.17.2` bytecode we ship:

- `QueueReleasingEmitterWrapper.cancel()` only does `isEmitterCanceled.set(true)`
- `QueueReleaseInterface.release()` is called **only** from `onComplete` / `onError`
- `ConnectionOperationQueueImpl$2$1.cancel()` only removes entries that have not started yet

So an already-running MTU operation keeps the connection's serial queue until its own ~30s native timeout. `discoverAllServicesAndCharacteristics` was issued immediately after and queued behind it, so it could never finish inside `BLE_GATT_SETUP_TIMEOUT_MS` (10s). `3s < 10s < 30s` made this deterministic.

The field log shows the sandwich, and a run from the day before on the same device shows the contrast: there the MTU request was allowed to retire natively (~30s) and discovery then completed in 18ms. Across that log, every acquire that logged `BLE MTU ready` finished GATT setup in 8-21ms; every acquire where the 3s bound fired took >10s. 16/16.

Two changes:
- skip the standalone refresh when the MTU-bearing connect already fell back — ble-plx applies `connectOptions.timeout` to the whole `establishConnection -> refreshGatt -> requestMtu` chain, so that fallback already means the peripheral will not finish an MTU exchange on this link
- if the bounded request is abandoned anyway, drop the link instead of discovering behind it

## 2. A cancelled native GATT setup skipped cleanup

ble-plx reports its own expired budget as `OperationCancelled`, not `OperationTimedOut`, so `abandonStalledConnection` never ran for the GATT stage. Both codes now count as abandoned native work. The connect stage keeps its own recovery ladder and instead cancels the connection before retrying on top of it, so it does not spend the manager-reset budget on a connect that is about to be retried anyway.

## 3. A sleeping Classic could not be woken

Its BLE co-processor keeps serving connect, GATT and MTU while the main MCU sits in the screensaver loop, where `msg_read_tiny` accepts only `Initialize` and the `*Ack` messages and drops everything else without a reply. `GetFeatures` (V1 probe) and `Ping` (V2 probe) are therefore both silently discarded — connection layer fully green, device screen black, nothing ever advances.

`Initialize` is the only message that breaks that loop, but `cf8416683` deliberately moved the probe off it because `fsm_msgInitialize` calls `session_startSession(NULL)`, which evicts a session slot and makes the active session a blank one — dropping a hidden-wallet session before Core can restore it via `Initialize {session_id}`.

So it stays a last resort rather than the probe: after a detection that got no answer on **any** protocol, the next attempt spends one `Initialize` wake on its freshly reconnected link before probing. A successful detection re-arms it. The wake is not cleared on teardown, because a failed detection tears the link down every poll and that would wake the device into a new session on every cycle.

## 4. Desktop: the routine post-call cancel tore down the kept-alive link

Separate root cause, separate platform. Desktop BLE paid a full cold reconnect on the operation *after* every successful hardware call. Field log 2026-09-12: of 10 physical teardowns, 2 were the main-process idle timer working as designed, 2 were the stale-bond error path, and **6 followed a `cancel API` carrying `requestQueues: length:0`** — nothing to cancel — on calls that had just returned `"success":true`.

`ElectronBleTransport.releaseNative()` used to skip the native disconnect unless `connectedDevices` still held the id. `releaseLogical()` clears that set, so `disconnect()` on an idle kept-alive link was a no-op and the main-process idle timer owned the teardown. In `77b76ac7be` (#905) that guard was hoisted and the disconnect became unconditional, which reconnected the teardown to hd-core's routine post-call path:

```
ServiceHardwareUI -> sdk.cancel -> interruptDevice (runs even with an empty queue)
  -> Device.interruptionFromUser `!acquired` branch -> deviceConnector.disconnect
  -> ElectronBleTransport.disconnect -> releaseNative -> nobleBle.disconnect
```

A/B against the same Pro 76D1 on a build that still had the guard (`app-2026-09-07.0.log`, pin `1.2.2-alpha.0`): `15:15:56` cancel, link stays up, `15:16:16.636 Keep-alive timeout ... "reason":"idle"` reclaims it 20s later. On `alpha.113`: `17:25:08` cancel, teardown 85ms later. Cancel-to-teardown coupling went from 0-3% on the August builds to 60%.

Cost of the avoidable cold connect, measured end to end (IPC CONNECT to `paired successfully`): **2.93s on Pro 76D1, 17.2s and 17.6s on Pro 2 B650** — the Pro 2 stops advertising for ~10.5s after a teardown, so its re-acquire burns a `connect-by-id` timeout plus three scan timeouts first.

### Why restoring the old guard alone is wrong

The guard was removed for a real defect, raised as P2 on #905: a native connect that never calls back leaves `connectedDevices` empty, so the Core 60s acquire deadline could not terminate it — the cold cleanup was a no-op.

That case and this regression **reach the same public `disconnect()` entry point**, and in both `connectedDevices.has(id)` is false, for opposite reasons: the link is stuck mid-acquire, or the link is idle and healthy under keep-alive. No argument the caller can pass separates them, because the caller is the same. They are only separable inside the transport, so this tracks in-flight acquires:

- `acquiringDevices` is populated at `acquire()` entry and cleared in `finally`
- the native disconnect is sent when the renderer either owns the link (`connectedDevices`) or is still trying to get it (`acquiringDevices`)
- `forceNative` stays available for the hard-teardown callers that must drop a link presumed dead (Protocol V2 link-fatal, V1 response timeout)

Net effect per caller: the deadline/abort case still disconnects (now proven through the public entry point), `REQUIRE_DISCONNECT` recovery still disconnects, and only the routine cancel on an idle link stops doing so. One extra change: an acquire that fails fast no longer sends a second disconnect, because `acquire()`'s own catch already sent one.

Also logs the `interruptionFromUser` disconnect. The connector drops the link silently, so a user-cancel teardown was indistinguishable in field logs from an idle keep-alive release or a device that left on its own; the chain above had to be reconstructed from the *absence* of an `Unsubscribing from notifications` line.

### Expected effect

Removes the reconnect at each operation boundary, when the next operation lands inside the 20s keep-alive window. Not a throughput change — MTU, pacing and device-side time are untouched. Arithmetic on the measurements above: one account on Pro 76D1 10.77s -> ~7.8s, on Pro 2 B650 27.3s -> ~10.1s. Pro 2 keeps its remaining exposure through `BLE_IDLE_DISCONNECT_MS` (20s): a pause longer than that still drops the link and still costs 17-26s to come back. Its ~10.5s advertising gap after a teardown is device-side and out of scope here.

## Tests

`packages/hd-transport-react-native`: 178 passing. Four new tests, each verified to fail when its fix is reverted:

- `drops the link when a stalled MTU negotiation is abandoned`
- `skips the MTU refresh after an MTU-bearing connect was cancelled`
- `cleans up a cancelled native GATT setup, not only a timed-out one`
- `spends one Initialize wake on the detection after a fully silent one`

`bounds a stalled MTU negotiation and continues protocol probing` asserted the behaviour that caused this bug and is replaced by the first of those.

`packages/hd-transport-web-device`: 35 passing. The desktop change rewrites the #905 test that locked the unconditional disconnect, rather than deleting it:

- `cancels a pending native connect acquire has not registered yet` — replaces `sends a native disconnect even before acquire marks the device connected`; drives a never-resolving `connect` and asserts through the public `disconnect()` instead of the private method, so it covers the actual deadline path
- `leaves a kept-alive link up when a released device is disconnected` — the keep-alive contract; verified to fail when the guard is reverted
- `forceNative disconnects a device the renderer no longer tracks` — keeps the explicit hard-teardown escape hatch covered

## Notes

- Firmware is out of scope here, so these are mitigations: the Pro 2 still does not answer the MTU exchange after a reconnect, and the Classic still drops probes while asleep. **Needs on-device regression before release.** For the desktop change the check is: after a successful `allNetworkGetAddress`, the `cancel API` line should be followed within 20s by `Keep-alive timeout ... "reason":"idle"`, and by no `manual-disconnect` that lacks a preceding `Unsubscribing from notifications`.
- Also drops a `void` operator that failed lint on the current branch tip (from `42498f699`), and a prettier violation in `packages/core/__tests__/search-devices.test.ts` inherited from `a9e4bf650`.
- **Merged the base in (`a115ca699`), and it drops `ad090fb3c`.** `alpha.114` is already published from the base branch, so without the merge `alpha.115` would have been a fork of it, not a superset — verified against the npm tarballs (`getRequestTasksIdByConnectId` present in `@onekeyfe/hd-core@1.2.2-alpha.114`, `protocolWakeAttempts` absent from `@onekeyfe/hd-transport-react-native@1.2.2-alpha.114`). Kept from the base: `662577083` (WebUSB search on the active request path, with its updated tests) and `7b38273b9` (changelog workflow). **Dropped: `ad090fb3c` "skip BLE MTU retry after cancelled connect"** — it and `bbc15ff1c` are concurrent fixes for the same Android MTU wedge written without knowledge of each other, and they disagree on the recovery action (tear the link down and rethrow, vs reconnect in place with fallback connect options). Resolved in favour of `bbc15ff1c` per the field-log analysis in section 1. `ad090fb3c` touched only the two files that resolution reverts, so nothing else of the base is affected. @wabicai — this needs your eyes before it lands.
- `agent:check --profile commit` passes everything except `@onekeyfe/hwk-trezor-connector test`, which fails identically on the pristine base branch in this checkout (`hwk-trezor-transport-common` is not linked in `node_modules`) and is unrelated to this change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/hd-transport-react-native/src/index.ts
```diff
@@ -1128,6 +1175,13 @@ export default class ReactNativeBleTransport {
           e.errorCode === BleErrorCode.OperationCancelled
         ) {
           Log?.debug('first try to reconnect without params');
+          mtuHandshakeRefused = true;
+          // The disposed chain can still own native connection state, and the retry
+          // would otherwise run on top of a half-open GATT client.
+          await this.runBestEffortNativeOperation(
```
originalix: P2: [Unbounded cancellation can wedge the reconnect fallback]

When the MTU-bearing connect is cancelled because its native chain is already stalled, this awaits `cancelDeviceConnection()` before attempting the fallback connect. That cancellation can be queued behind the same stalled native operation and never settle, so `acquire()` hangs instead of reaching the fallback or resetting the BLE manager.

Run this cleanup through `runNativeTeardown` (and do the same for the `disconnectedDevice.cancelConnection()` branch below) so the existing teardown timeout can reset the manager and allow the retry path to settle.

packages/hd-transport-react-native/src/index.ts
```diff
@@ -2485,6 +2550,47 @@ export default class ReactNativeBleTransport {
...
+    if (this.protocolWakeAttempts.has(uuid)) return;
+    if (!transportCache[uuid]) return;
+    this.protocolWakeAttempts.add(uuid);
+
+    Log?.debug('[ReactNativeBleTransport] sending Protocol V1 Initialize wake', {
+      connectIdSuffix: uuid.slice(-8),
+    });
+    try {
+      this.probingProtocols.set(uuid, 'V1');
+      await this.callProtocolV1(uuid, 'Initialize', {}, { timeoutMs: PROTOCOL_PROBE_TIMEOUT_MS });
```
originalix, line 2710: P1: [Initialize wake self-deadlocks React Native BLE acquisition]

After a fully silent detection, this new `Initialize` wake is allowed to receive no reply. Its timeout is not classified as a probe timeout in `callProtocolV1` (only `GetFeatures` is), so it awaits the public `disconnect()` path.

`acquire()` already owns `runLifecycleOperation()` for this UUID, and `disconnect()` queues behind that operation. The wake therefore waits for the acquire to finish while the acquire waits for the wake to return, leaving the sleep/wake recovery stuck indefinitely.

Treat this wake timeout as a non-destructive probe timeout so the subsequent `GetFeatures` probe can run, and add a regression test where `Initialize` never responds.

PR #939, opened 2026-09-15
docs: document EVM clear signing and Safe transactions

wabicai:
The developer portal did not explain how to integrate EVM Clear Signing or distinguish parsed EIP-712 and Safe owner signatures from hash signing. Update the English and Chinese signer guides and API references with the supported request formats, device review expectations, fallback limits, and Safe Tx examples.

Correct transaction `chainId` placement, required typed-data options, hash preparation, serialization examples, and guide links. The transaction reference explicitly covers Legacy and EIP-1559; its serialization example is limited to EIP-1559 (Type 2). Align returned fields and error-code types with the SDK. Message verification examples add the `0x` prefix to returned signatures, exclude `EIP712Domain` from ethers message types, and check the recovered address. Document the Touch/Pro firmware 4.2.0 minimum for hash signing. Changes are limited to eight documentation pages.

Validation:
- Compiled the changed MDX pages and checked their local links; the six pages updated in the latest commit passed 65 link checks.
- Checked TypeScript syntax in 34 snippets; type-checked and executed 20 complete examples with mock SDK responses, including approval calldata, typed-data hashes, signature verification, and transaction serialization.
- Passed 12 focused verification cases across both languages and personal_sign/EIP-712: success, rejected SDK request, and mismatched signer address. Synthetic signatures were used without private keys or device calls.
- `yarn agent:check --profile commit` and package version checks passed.
- `yarn agent:check --profile pr` stopped during full-repository lint: Node reached the configured 8 GB heap limit. The full-repository test/build stages were not reached.
- Hardware screen behavior was checked against source; no physical-device screen validation was performed.

PR #941, opened 2026-09-15
docs: align SDK developer guides with current API contracts

wabicai:
Hardware SDK integration guides contained incorrect API names, missing parameters, inaccurate result types, and event/transport guidance that could leave integrations blocked or using the wrong wallet. Align the English and Chinese developer portal with the current SDK implementation after the Clear Signing documentation update in #939.

Reviewed all 262 hardware-sdk pages and corrected 234 (117 per language). Most changes are repeated error-code/path type corrections. Focused guide changes fix initialization and wallet guards, device/UI event subscriptions, cancellation results, raw BLE chunk handling for V1/V2, BTC/Solana examples, Cardano governance parameters, device settings, firmware-check nullability, and TON proof timestamps. Repair the eight broken internal anchors.

The XRP signing declaration, Nostr public-key alias, and Benfen signing result declaration still differ from runtime. Their reference pages now explicitly describe that limitation and the runtime contract. This PR changes documentation only.

Validation:
- All 262 MDX pages compile; 849 internal links pass; executable-fragment syntax and SDK method-name scans pass.
- Checked 597 SDK-call snippets against current CoreApi types: 419 have no diagnostics; the rest depend on surrounding/host variables, except the two documented XRP declaration errors (one per language).
- 70 documented lifecycle, wallet-selection and UI-event cases pass across both languages using mock SDK responses.
- 20 complete EVM examples type-check and execute; 12 personal_sign/EIP-712 signature-verification cases pass.
- `yarn agent:check --profile commit` and package-version checks pass.
- `yarn agent:check --profile pr` reaches full-repository lint, which exhausts the configured 8 GB Node heap locally; full test/build stages were not reached. No physical-device or native-platform validation was performed.


packages/connect-examples/developer-portal/content/en/hardware-sdk/basic-api/get-passphrase-state.mdx
```diff
@@ -47,7 +47,7 @@ Result
 ```
 {
     success: true,
-    payload: string
+    payload: string | null | undefined
```
originalix, line 50: P2: [Documented V2 passphrase-state result is incorrect]

When an integrator calls `getPassphraseState` for a V2 standard-wallet flow, this page now tells them to expect `null`. The exported method type is `Response<string | undefined>`, and the V2 run path returns the session `passphraseState`; it does not produce `null`.

`null` is the standard-wallet value returned by `openWalletSession`, so this conflates two APIs and can make callers reject a valid result or model the wallet binding incorrectly. Please document `getPassphraseState` as `string | undefined` here and in the Chinese page, and keep the `null` explanation on the `openWalletSession({ mode: 'standard' })` contract.

ByteZhang1024 (approved):

